PR #1, opened 2017-04-12
Rewrite associative memory.

jgosmann:
(This PR has been moved here from nengo/nengo#1253. The discussion from that PR has not been copied over.)

**Motivation and context:**
This PR introduces a rewritten associative memory for the `new-spa`.

Compared to previous implementations where the `AssociativeMemory` implemented all possible variants, I split up different possible implementations.

The core networks that can be used to implement something like an associative memory are found in `selection.py`. These are a simple array of thresholding ensembles (`ThresholdingArray`), a basic WTA (`WTA`) with lateral inhibition, and the independent accumulator (`IA`) network submitted to CogSci. All these core networks are standard Nengo networks because they don't operate on semantic pointers, but each dimension represents a potential choice. (I also wanted to repeatedly use such networks in situations where I didn't have semantic pointers.) This also leads to the question: **Do we want those networks potentially in core Nengo? Or do we keep them in nengo_spa?**

The `AssociativeMemory` is basically a wrapper around such a core network that adds the projections from and to the Semantic Pointer space and allows to use the network with other SPA modules. We pass only the class of the core network + kwargs to the AssociativeMemory to avoid the user to have to pass `add_to_container=False`. Also, for better usability for the end-user there are classes `ThresholdingAssocMem`, `WtaAssocMem`, and `IaAssocMem`.

Further comments:
* I am not setting a default value for the threshold. To me it seems better to explicitly force the user to specify the threshold.
* By default the identity function is applied to the outputs of `ThresholdingArray` and `WTA`. A step function might be more appropriate in some cases, but is easily added by providing the `function=lambda x: x > threshold` argument.
* I haven't implemented an `inhibitable` flag. One still has the option to create inhibitory connections to the core network and other SPA modules don't provide this option either (actually I would find it more useful on `spa.State` then on the `AssociativeMemory`). I plan to make a PR to core Nengo with a function that inhbits all ensembles in a network. It's a functionality I commonly need, even outside of SPA. That function could than also be used for all SPA modules.
* I haven't implemented the cleanup with double inhibition. It never made it into the Nengo master (but exists in @xchoo's `assoc_mem_v3` branch). I'm not sure whether we want to include it. It reduces the noise on the output in certain circumstances, but I haven't missed the functionality and it could always be added later.
* There is no non-SPA associative memory that uses vectors anymore. If you need that, you can use one of the core networks and set the transforms on the input and output connection yourself. It doesn't seem necessary to have just a network for that and chances are high if someone is using vectors, the SPA associative memory with a proper vocab can be used.

This PR supersedes #991.

**How has this been tested?**
Updated and adapted existing tests.

**Checklist:**

- [x] I have updated the documentation accordingly. [The associative_memory example probably needs to be updated, but all the SPA examples need to be updated at some point which will be a separete effort.]

**Still to do:**
- [x] documentation
- [x] drift diffusion network
- ~~cleanup with double inhibition?~~

### DIFF
--- a/examples/associative_memory.ipynb
+++ b/examples/associative_memory.ipynb
@@ -1,9 +1,9 @@
 {
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "nengodev",
    "language": "python",
-   "name": "python3"
+   "name": "nengodev"
   },
   "language_info": {
    "codemirror_mode": {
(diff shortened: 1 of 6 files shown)


jgosmann:
Abbreviations with camel case are always a bit weird and I don't really care which variant we are picking. So I changed that und removed the `Array` part from `Thresholding`. Good to merge now?

PR #4, opened 2017-04-13
Implement spaopt.

jgosmann:
**Motivation and context:**

This improves the accuracy of most SPA components by setting intercept and evaluation point distributions. For details see
[The Influence of the Intercept Distribution on Representation in the NEF](http://localhost:8890/notebooks/The%20influence%20of%20the%20intercept%20distribution%20on%20representation%20in%20the%20NEF.ipynb). Note, that this achieves accuracies comparable or better to what I presented [in this paper](http://journals.plos.org/plosone/article?id=10.1371/journal.pone.0149928), but doesn't require any prior optimization step.

none

**How has this been tested?**
Tests still pass, in some tests the tolerances needed to be adjusted. This is either an issue of the tolerances being too tight or because many of the tests use rate neurons. The spaopt optimizations are for spiking neurons. For rate neurons with the default solver regularization they should still slightly improve things on average, but the overall distortion changes, i.e. in some parts of the space the representation gets worse while improving for other parts. I think it is fine to optimize for the spiking case as the accuracy with rate neurons will not be worse and for optimal performance with rate neurons the user would have to change the solver regularization too, so this adds just two more parameters that need to be adjusted for optimal performance with rate neurons.

**Types of changes:**

**Checklist:**

- [n/a] I have updated the documentation accordingly. (Will happen later.)
- [n/a] I have included a changelog entry. (prior to first release)

**Still to do:**

- [x] test represent identity
- [x] In some tests the error bounds had to be increased. This might have been too tight before or it might be an artifact of too low dimensionality, but it should be verified that things are not actually worse on average.

### DIFF
--- a/nengo_spa/state.py
+++ b/nengo_spa/state.py
@@ -147,7 +147,7 @@ class State(Network):
     feedback_synapse = NumberParam(
         'feedback_synapse', default=.1, readonly=True)
     represent_identity = BoolParam(
-        'represent_identity', default=False, readonly=True)
+        'represent_identity', default=True, readonly=True)
 
     def __init__(self, vocab=Default, subdimensions=Default,
                  neurons_per_dimension=Default, feedback=Default,
--- a/nengo_spa/state.py
+++ b/nengo_spa/state.py
@@ -1,5 +1,3 @@
-import numpy as np
-
 import nengo
 from nengo.exceptions import ValidationError
 from nengo.networks.ensemblearray import EnsembleArray
--- a/nengo_spa/state.py
+++ b/nengo_spa/state.py
@@ -65,8 +65,8 @@ class State(Network):
                 self.neurons_per_dimension * self.subdimensions,
                 dimensions // self.subdimensions,
                 ens_dimensions=self.subdimensions,
-                radius=np.sqrt(float(self.subdimensions) / dimensions),
-                #  TODO radius
+                eval_points=nengo.dists.CosineSimilarity(dimensions + 2),
+                intercepts=nengo.dists.CosineSimilarity(dimensions + 2),
                 label='state')
 
             if self.feedback is not None and self.feedback != 0.0:
--- a/nengo_spa/state.py
+++ b/nengo_spa/state.py
@@ -1,9 +1,115 @@
+import warnings
+
 import nengo
 from nengo.exceptions import ValidationError
 from nengo.networks.ensemblearray import EnsembleArray
-from nengo.params import Default, IntParam, NumberParam
+from nengo.utils.network import with_self
+from nengo.params import BoolParam, Default, IntParam, NumberParam
 from nengo_spa.network import Network
 from nengo_spa.vocab import VocabularyOrDimParam
+
+
+class IdentityEnsembleArray(nengo.Network):
+    """An ensemble array optimized for representing the identity vector for
+    circular convolution.
+
+    The ensemble array will use ensembles with *subdimensions* dimensions,
+    except for the first *subdimensions* dimensions. These will be split into
+    a one-dimensional ensemble for the first dimension and a *subdimensions-1*
+    dimensional ensemble.
+
+    Parameters
+    ----------
+    neurons_per_dimension : int
+        Neurons per dimension.
+    dimensions : int
+        Total number of dimensions. Must be a multiple of *subdimensions*.
+    subdimensions : int
+        Maximum number of dimensions per ensemble.
+    kwargs : dict
+        Arguments to pass through to the `nengo.Network` constructor.
+
+    Attributes
+    ----------
+    input : nengo.Node
+        Input node.
+    output : nengo.Node
+        Output node.
+    """
+    def __init__(
+            self, neurons_per_dimension, dimensions, subdimensions, **kwargs):
+        super(IdentityEnsembleArray, self).__init__(**kwargs)
+
+        self.neurons_per_dimension = neurons_per_dimension
+        self.dimensions = dimensions
+        self.subdimensions = subdimensions
+
+        self.neuron_input = None
+
+        cos_sim_dist = nengo.dists.CosineSimilarity(dimensions + 2)
+        with self:
+            self.input = nengo.Node(size_in=dimensions)
+            self.output = nengo.Node(size_in=dimensions)
+
+            first = nengo.Ensemble(neurons_per_dimension, 1)
+            nengo.Connection(self.input[0], first, synapse=None)
+            nengo.Connection(first, self.output[0], synapse=None)
+
+            if subdimensions > 1:
+                second = nengo.Ensemble(
+                    neurons_per_dimension * (subdimensions - 1),
+                    subdimensions - 1,
+                    eval_points=cos_sim_dist, intercepts=cos_sim_dist)
+                nengo.Connection(
+                    self.input[1:subdimensions], second, synapse=None)
+                nengo.Connection(
+                    second, self.output[1:subdimensions], synapse=None)
+
+            if dimensions > subdimensions:
+                remainder = nengo.networks.EnsembleArray(
+                    neurons_per_dimension * subdimensions,
+                    dimensions // subdimensions - 1, subdimensions,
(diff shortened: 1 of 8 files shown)


jgosmann:
@xchoo @tcstewar This introduce a `represent_identity` option for `spa.State` to allow the representation of the identity vector (which cannot be represented very well with the default `CosineSimilarity` intercept and eval point distributions for spaopt). I wonder whether the default should be `True` or `False`. I usually do not encounter the identity vector in the things I'm doing, but it might be better to err on the side that it might occur?

tcstewar:
> I wonder whether the default should be True or False.

Hmm...  interesting question.  I think I'd go with defaulting to ```False```, given how uncommon it is.  But that could also be that it's never really been something I've needed.

jgosmann:
Dev meeting decision: set default for `represent_identitiy` to `False`

xchoo:
Whoopsy! I should've attended the dev meeting, but I didn't realize this was going to be brought up during the most recent dev meeting.

I would disagree with setting the default to `False`. In my experience, if folks do not work through the semantic pointer math, and explicitly work out the products of each step of the math, it is very easy to overlook needing this requirement when building a SPA network. I believe that pretty much any time you perform a transformation calculation (see Dan's raven's stuff), the identity transform will pop up in the results somewhere. 

While it is true that for non-transform cases, the identity transform doesn't appear as often, I'd still advocate setting the default to `True` because:

- You never know when the identity transform pops up and needs to be included in the network.
- Debugging a missing identity transform is very difficult unless you know exactly what you are looking for. (case in point: it took me 2+ years to realize this was why the Spaun network was not doing the induction task as accurately as it was supposed to)
- Because it only affects 1 dimension of the vector, setting the default to `True` doesn't impact the representational accuracy of the <insert module here>.

jgosmann:
> Because it only affects 1 dimension of the vector, setting the default to True doesn't impact the representational accuracy of the .

This is not correct. It affects one ensemble which is by default 16 dimensions, but could be all of the dimensions.

Though, it just occured to me that we could construct a distribution where only the first dimension is uniformly sampled. In that case it is indeed only one dimension that is affected.

xchoo:
> This is not correct. It affects one ensemble which is by default 16 dimensions, but could be all of the dimensions.

Ah. Right. Perhaps I should say that it *should* only affect the first dimension network. (If it doesn't, that should be addressed. 😄 )

jgosmann:
> Though, it just occured to me that we could construct a distribution where only the first dimension is uniformly sampled. In that case it is indeed only one dimension that is affected.

While this works for evaluation points (where we actually sample in the d-dimensional space), it currently cannot be done for the intercepts. These are just one dimensional samples that are projected into the d-dimensional space with the encoder. So the distribution from which the intercept of each neuron is picked would have to depend on the neuron's encoder.

Another approach would be to create a one-dimensional ensemble for the first dimension, but I'm not sure what potential problems that might introduce.

jgosmann:
This should be ready for review now (I updated the PR description). The default for `represent_identity` is `False` at the moment, but that can still be changed after this PR has been merged. Maybe it is best, if you, @tcstewar and @xchoo, directly discuss this? You two probably have the strongest opinions. For convenience let me summarize the different ways of implementing the identitiy representation:
1. Setting `eval_points = UniformHypersphere()` and `intercepts = Uniform(-1., 1.)` (current implementation). This affects all subdimensions in the first ensemble, which is 16 by default. The higher the subdimensions parameter, the more the representation for random vectors is worsned by `represent_identity = True`.
2. Setting `eval_points = UniformHypersphere()` and leave `intercepts = CosineSimilarity(d + 2)`. This affects all subdimensions in the first ensemble, which is 16 by default. The higher the subdimensions parameter, the more this approaches `represent_identity = False` (while still being able to represent the identity), so this is better than 1 in this respect. This approach has the highest error for representing random vectors for intermediate values of subdimensions (like 16), but is probably not worse than 1. The more concentrated intercepts seem to not matter a lot for representing the identity.
3. Like to but constructing a representation for `eval_points` where the first dimension is `Uniform(-1., 1.)` and the rest `CosineSimilarity(d + 2)`. Surprisingly, this leads to a worse representation for the identity (it is overshot by up to 0.2). Also, random vectors are represented less accurately.
4. Splitting the first dimension of into a separate ensemble. I suppose this is mostly implemenattion-wise annoying because it gives ensembles of different sizes. But at least it should not noticeably decrease the accuracy for random vectors.

Option 1 to 3 all worsen the accuracy for representing random vectors.
index
This plot shows error distributions for representing 64-dimensional random unit vectors, the tuple on the left states the distributions/optimizations used and the number of subdimensions. `spaopt` is the old spaopt, `cosine` is the new spaopt (implemented in this PR without `represent_identity`), `cosine_identity` corresponds to option 2, and `cosine_identity2` corresponds to option 3. The most interesting lines are probably ('cosine_identity', 16) and ('cosine_identity2', 16) compared to ('cosine', 16) (which is more accurate) and ('spaopt', 16) (which is about the same).

jgosmann:
I think my favoured option is 4 and making `represent_identity` the default. With option 4 the accuracy for random vectors is only marginally reduced and it is possible to accurately represent the identity to. If you need the last bit of accuracy and don't need to represent the identity, you can still set it to false (thanks to the config system even globally). Same if you require the simpler ensemble array structure with equally sized ensembles.

jgosmann:
I added commits to implement the fourth proposed option.

nengo_spa/state.py
```diff
@@ -1,13 +1,110 @@
...
+
+        Returns
+        -------
+        nengo.Node
+            The added node.
+        """
+        if self.neuron_input is not None:
+            warnings.warn("neuron_input already exists. Returning.")
+            return self.neuron_input
+
```
tcstewar: this doesn't have the check for `nengo.Direct` that is in the standard `EnsembleArray` class

jgosmann: Yeah, it seemed unecessary/redundant to me. The user can change the neuron type after adding the neuron input, so the check would have not effect in that case. And in any case, an error should be raised on build time.

jgosmann: But I'm happy to add it if you insist.

tcstewar: I think for consistency with `EnsembleArray` we should have it.  It's so similar to the same function in `EnsembleArray` that it just seems odd to have that one difference.  But no need to go replicate all those other functions.

jgosmann: k, though those networks might go out of sync anyways
At some point (depending on how things turn out) it might be possible to remove `add_neuron_input` and use `create_inhibitory`.

jgosmann:
It might be worth to mention that changing the `represent_identity` default globally for a single model is a one-liner:
```python
with nengo_spa.Network() as model:
    model.config[nengo_spa.State].represent_identity = False
    # ...
```

nengo_spa/state.py
jgosmann: Added a commit, can you give a quick check? (It's slightly different from `EnsembleArray`. It checks all ensembles instead just the first one.)

nengo_spa/state.py
```diff
@@ -91,6 +91,13 @@ def add_neuron_input(self):
             warnings.warn("neuron_input already exists. Returning.")
             return self.neuron_input
 
+        if any(isinstance(e.neuron_type, nengo.Direct)
+               for e in self.all_ensembles):
+            raise ValidationError(
+                "Ensembles use Direct neuron type. "
+                "Cannot getneuron output from Direct neurons.",
```
tcstewar: That error message is wrong...   you want "Cannot give neuron input to Direct neurons." (which is the same as in EnsembleArray)

nengo_spa/state.py
tcstewar: I'm good with you checking all the ensembles instead of the first one, but the text of the error message is wrong.  :)

nengo_spa/state.py
jgosmann: And that's why do PR reviews ... copied from the wrong place. Should be fixed now.

PR #27, opened 2017-05-12
Overhaul input system

jgosmann:
**Motivation and context:**
This changes how SPA inputs work by making them more similar to Nengo nodes. That has two main advantages:

* The user doesn't need to learn a new syntax. Inputs are created as normal SPA module an can be connected like any other SPA module in action rules.
* It simplifies the code because it is based on the standard module code.

This makes it slightly more effor to define many function inputs as for each one a new module has to be created and connected. But as standard Nengo requires the same amount of effort when using nodes, I believe this to be tolerable. Non-function inputs (static inputs) can be provided within the action rules. There is no need to create an extra module for that.

Furthermore, `Input` has been renamed to `Encode` (a module that encodes a semantic pointer expression into neural activities). In addition, the modules `Decode` and `Transcode` have been added. The difference is in whether the input and output have vocabularies associated:

* `Encode` has a vocabulary associated with its output. That means the provided function may return a string that gets parsed with that vocabulary, a `SemanticPointer` (used as is) or a NumPy array (also used as is). The function is also allowed to accept an input array that is provided by a Nengo connection into `Encode.input`. This input has no vocabulary associated with it and the size is determined via `size_in`.
* `Decode` has a vocabulary associated with its input. The provided function must take three arguments (instead of the usual one or two): a float for the timestep, a `SemanticPointer` (the current input) and the input vocabulary. The function may return an NumPy array as output. No vocabulary is associated with it.
* `Transcode` has an input and output vocabulary. The function arguments are as for `Decode` and the expected return value may be a string or `SemanticPointer` as for `Encode`. Using `None` as function essentially creates a pass-through node, but with vocabularies which might come in handy for plotting in the GUI.

Another way to express the roles of these classes: `Encode` is used to translate from non-SPA vectors (if any are provided as input, otherwise just time-dependent) to semantic pointers, `Decode` reads semantic pointers and can translate to non-SPA vectors, `Transcode` uses some arbitrary function to turn semantic pointers in one vocabulary into vectors of another vocabulary.

`Encode` does also support static strings in addition to functions (like Nengo nodes support static values, but using action rules for this kind of input is usually more convenient). I thought about implementing support for processes, but this would require to implement a SPA specific process class which seems like a significant amount of work without the need being clear to me. So I left this for the future.

This PR would address #11.

none

**How has this been tested?**
Added tests.

**Types of changes:**

**Checklist:**

**Still to do:**
- [x] Test pass-through


nengo_spa/modules/decode.py
```diff
@@ -0,0 +1,39 @@
...
+    vocab = VocabularyOrDimParam(
+        'vocab', optional=False, default=None, readonly=True)
+    size_out = IntParam(
+        'size_out', low=0, optional=True, default=None, readonly=True)
+
+    def __init__(
+            self, function=Default, vocab=Default, size_out=Default, **kwargs):
+        super(Decode, self).__init__(**kwargs)
+
+        self.vocab = vocab  # needs to be set before function
```
Seanny123: Why does the vocab have to be set before the function?

nengo_spa/modules/tests/test_encode.py
```diff
@@ -0,0 +1,82 @@
...
+
+        model.input = spa.Encode(input, vocab=16)
+        spa.Actions('buffer = input').build()
+
+        p = nengo.Probe(model.buffer.output, synapse=0.03)
+
+    with Simulator(model) as sim:
+        sim.run(0.3)
+
+    input, vocab = model.get_network_input('buffer')
```
Seanny123: Maybe don't use the reserved keyword `input`?

nengo_spa/modules/decode.py
jgosmann: The validation of the `TranscodeFunctionParam` will check the vocabulary dimensionality.

nengo_spa/modules/tests/test_encode.py
jgosmann: These were copied from the old `Input` tests. But you're right that it might make sense to clean such things up when already touching the code. I will take a look at it.

nengo_spa/modules/decode.py
Seanny123: Might want to include that detail in the comment.

jgosmann:
I addressed @Seanny123 comments.

jgosmann:
@tcstewar, do you want to take a look at this too (because it's a significant change to the API)?

jgosmann:
If I don't get any response, I will merge this before June 23 (that's the date I will present nengo_spa in the lab meeting).

Issue #8, opened 2017-05-09
Replace associative memory input_keys and output_keys with single mapping parameter.


jgosmann:
This came up originally in nengo/nengo#891 and [this comment](https://github.com/nengo/nengo/issues/891#issuecomment-153885809), partially reproduced here, describes the actions to take:

Replace `AssociativeMemory(input_vocab, output_vocab=None, input_keys=None, output_keys=None, ...)` with `AssociativeMemory(input_vocab, output_vocab=None, mapping=None, ...)`.
- This does not allow to pass in a set of keys to cleanup a subset of a vocab, but this can still be done by using `Vocabulary.create_subset()` and passing that in (which is what I did so far anyways because `input_keys` is undocumented).
- If `output_vocab` is set, but `mapping` is not, raise an exception to force the user to be explicit about what should happen.
- If `mapping='by-key'` (or something similar) do the mapping according to the keys (transform-style memory).
- To do a hetero-associative memory one passes in a dictionary with a mapping from keys in `input_vocab` to keys in `output_vocab`. (To me this seems to make it more clear how the mapping happens than two lists named `input_keys` and `output_keys`.)

There was some concern about unintuitive behaviour of the `Vocabulary.create_subset` method, but these problems should be resolved in `nengo_spa`.

PR #29, opened 2017-05-19
Implement mapping argument for associative memories.

jgosmann:
**Motivation and context:**
The interaction of the `output_vocab`, `input_keys`, and `output_keys` arguments for associative memories was a bit confusing. This PR hopes to make things clearer by replacing `input_keys` and `output_keys` with a single `mapping` argument as proposed in #8.

none

**How has this been tested?**
With existing tests, most use `mapping=None`, but there is one test using `mapping='by-key'`, and one test using an explicit mapping which covers all options.

**Types of changes:**

**Checklist:**


jgosmann:
@ikajic, you opened the original issue nengo/nengo#891 relating to this. Would you like to do the code review?

nengo_spa/modules/assoc_mem.py
```diff
@@ -54,27 +58,28 @@ class AssociativeMemory(Network):
...
+                "vocabulary is given.", attr='mapping', obj=self)
         self.input_vocab = input_vocab
         self.output_vocab = output_vocab
 
-        if input_keys is None:
-            input_keys = self.input_vocab.keys()
-            input_vectors = self.input_vocab.vectors
-        else:
-            input_vectors = [input_vocab.parse(key).v for key in input_keys]
+        if mapping is None or mapping == 'by-key':
```
Seanny123: Should this throw a validation error if an invalid string is passed, informing people of the correct eligible strings? Should an error also be thrown if an invalid argument type is passed?

jgosmann: In the string case that might make sense, but I probably would not check the type. That is always a bit problematic with duck typing because you don't really want to check for the type, but whether the object behaves the right way (in this case it needs to have a `keys` method returning an iterable and support indexing with the elements).

Seanny123: I totally agree that checking the right string is enough and anything else would probably be annoying.

jgosmann: I addressed this comment.

ikajic:
Before providing the review I wanted to test this change so I created a new virtenv with `nengo_spa` installation and the requirements. For the following code:

```
d = 64
v1 = spa.Vocabulary(dimensions=d)
v1.parse('A')
```
I am getting `SpaParseError: Semantic pointers must start with a capital letter.`. Have you encountered this before?

jgosmann:
In `nengo_spa` `parse` will only accept names of semantic pointers that are already in the vocabulary. To add pointers to the vocabulary, the `populate` method is used (e.g. `v1.populate('A')`) which also supports some fancy new features. An alternate method is to restore the old behaviour by providing the `strict=False` argument to `Vocabulary`, in that case `parse` will silently add unknown pointers to the vocabulary.

The reason to be more strict about when semantic pointers get added to a vocabulary is that there were cases where pointers would accidentally get added to the vocabularies for associative memories and then those memories would clean up to items they should not clean up to. Also makes it more likely to spot mistyped semantic pointer names.

ikajic:
> In nengo_spa parse will only accept names of semantic pointers that are already in the vocabulary

That makes sense. The error I got was a bit confusing, so in case you don't already have it on your to-do list, you might want to update it so that it reflects the problem

ikajic:
Why does this throw an error?
```ipython
In [45]: v1
Out[45]: <nengo_spa.vocab.Vocabulary at 0x7fd2cbb00190>

In [46]: v2
Out[46]: <nengo_spa.vocab.Vocabulary at 0x7fd2cbabbd90>

In [47]: v1.keys()
Out[47]: ['A', 'B']

In [48]: v2.keys()
Out[48]: ['C', 'D']


```

jgosmann:
I think your code snippet is incomplete? Not sure what error you mean.

ikajic:
> I think your code snippet is incomplete? Not sure what error you mean.

I updated the post, sorry.

jgosmann:
`v2` does not have the keys `A` and `B`, so a mapping by key is not possible. I created #34 regarding the misleading error message.

ikajic:
> v2 does not have the keys A and B, so a mapping by key is not possible. I created #34 regarding the misleading error message.

Would this be the same error message though? I think in #34 the error should inform the user about a SP not being in the vocabulary (which is independent of AM), while in the snippet posted above the error message refers to something related to the AM module (and it would basically say something like what you just explained).

(I am guessing error messages are not essential part of the dev process at this stage, but they can mislead when reviewing and testing the code, so that's why I raised this up.)

jgosmann:
I'd say that it is the same error from a technical point. In both cases a requested Semantic Pointer is not part of the the vocabulary (in this case `A` and `B` are not found in `v2`). From a user perspective using the same error message in both cases might not make it clear what the problem in the associative memory case is. So definitely something to keep in mind and look into again once #34 is resolved.

nengo_spa/modules/assoc_mem.py
```diff
@@ -31,10 +34,12 @@ class AssociativeMemory(Network):
...
         None, the associative memory will act like an autoassociative memory
         (cleanup memory).
-    input_keys : list, optional (Default: None)
-        A list of strings that correspond to the input vectors.
-    output_keys : list, optional (Default: None)
-        A list of strings that correspond to the output vectors.
+    mapping: dict or str, optional (Default: None)
+        A dictionary that defines the mapping from Semantic Pointers in the
+        input vocabulary to Semantic Pointers in the output vocabulary. If set
+        to the string ``'by-key'``, the mapping will be done based on the keys
```
ikajic: I'm thinking whether it's worth changing 'by-key' to just 'key'. I'd say it's a bit easier to write. This is just a suggestion so you can also leave 'by-key' if you like that better

Seanny123:
I'd personally prefer "by-key" because it makes mapping="by-key" sound more
like natural English.

PR #31, opened 2017-05-19
Remove deprecated stuff.

jgosmann:
**Motivation and context:**
This removes a bunch of code that only existed to support legacy syntax and API from the old `nengo.spa`. One particular change is to not track the SPA networks manually anymore, but just look at the network's attribute (each module has to be assigned as an attribute anyways to be accesible in action rules).

none

**How has this been tested?**
Existing tests still pass.

**Checklist:**

### DIFF
--- a/nengo_spa/network.py
+++ b/nengo_spa/network.py
@@ -5,7 +5,6 @@ from nengo.config import Config, SupportDefaultsMixin
 from nengo_spa.exceptions import SpaNetworkError
 from nengo_spa.modules.input import Input
 from nengo_spa.vocab import VocabularyMap, VocabularyMapParam
-import nengo_spa
 
 
 def get_current_spa_network():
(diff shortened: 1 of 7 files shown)


nengo_spa/network.py
```diff
@@ -166,9 +166,6 @@ def get_network_input(self, name):
         except KeyError:
             raise SpaNetworkError("Could not find network input %r." % name)
 
-    def get_input_vocab(self, name):
```
Seanny123: Just to make sure I'm understanding this change, only the associative memory has an input vocab, so now all the networks have vocab attributes instead? Consequently, these accessors aren't required?

nengo_spa/network.py
```diff
@@ -187,18 +186,3 @@ def get_network_output(self, name):
                     raise KeyError
         except KeyError:
             raise SpaNetworkError("Could not find network output %r." % name)
-
-    def similarity(self, data, probe, vocab=None):
```
Seanny123: Note to other reviewer, this has been replaced by `examine.similarity`.

nengo_spa/network.py
jgosmann: If you require the vocab of a module, it is better to use the `vocab`, `input_vocab`, or `output_vocab` parameters according to the module. These functions query the vocabularies associated with specific inputs or outputs in the `self.inputs` and `self.outputs` dictionaries which is more useful for internal use when connecting things up with action rules, but apparently not that useful anymore as these functions weren't used anymore.

Issue #34, opened 2017-05-25
Vocabulary.parse gives misleading exception in strict mode


jgosmann:
```python
d = 64
v1 = spa.Vocabulary(dimensions=d)
v1.parse('A')
```
will give `SpaParseError: Semantic pointers must start with a capital letter.` which is misleading because the problem here isn't a non-capital letter, but that the vocabulary is in strict mode and `A` is not yet in the vocabulary.

PR #35, opened 2017-05-31
Improve error messages

### LINKED ISSUE #32: Better error handling
The `get_spa_network`, `get_network_input`, and `get_network_output` methods on `spa.Network` contain code that is intended to produce an error message that shows the full name, but it doesn't work because the `KeyError` is immediately catched in converted ot the `SpaNetworkError` instead of being reraised to the outermost invocation. This should be fixed. In addition, it would be could to catch the error in the `Actions` object and attach information about which exact rule failed.

jgosmann:
**Motivation and context:**
This PR improves some error messages to fix #32 and #34. As this touches `spa.Network.get_network_input` and `spa.Network.get_network_output`, I used this to also refactor these functions to reduce code duplication and make the implementation slightly simpler.

There is one exception, when a `spa.Network` attribute gets reassigned, that I completely removed because it seems to be unnecessary. I think the reassignment was only a problem in the legacy SPA, but I'm not entirely sure for the initial reason to introduce this exception. Tagging @Seanny123 and @tcstewar who might remember more details. Also, see description of commit d82bab8940fcbdc8ac9de3796171710ccc29886c.

In #29, @ikajic noted that an associative memory gives a misleading exception when the output vocabulary is missing keys. The general parse error given by vocabularies has been fixed in this PR, but currently the associative memory does not add any further detail to it. For example, for code like this
```python
spa.ThresholdingAssocMem(threshold=0.2, input_vocab=v1, output_vocab=v2, mapping='by-key')
```
the following exception might be produced:
```
SpaParseError: Error parsing expression 'A' with 32-dimensional vocab at 0x7f954d053f28: name 'A' is not defined
```
This is definitely an improvement, but is it good enough? If not how can this error message be made more helpful?

none

**How has this been tested?**
Unit tests still pass. Tested the parsing exception for `Vocabulary` manually because it is a bit of a hassle to verify the message string in a unit test.

I'd recommend doing this on a commit by commit basis.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,13 @@ Release History
   (`#30 <https://github.com/nengo/nengo/pull/30>`_,
   `#23 <https://github.com/nengo/nengo/issues/23>`_)
 
+**Fixed**
+
+- Improved a number of error messages.
+  (`#35 <https://github.com/nengo/nengo/pull/35>`_,
+  `#32 <https://github.com/nengo/nengo/issues/32>`_,
+  `#34 <https://github.com/nengo/nengo/issues/34>`_)
+
 
 0.1.1 (May 19, 2017)
 ====================
--- a/nengo_spa/ast.py
+++ b/nengo_spa/ast.py
@@ -87,7 +87,7 @@ from nengo_spa.modules.bind import Bind
 from nengo_spa.modules.compare import Compare
 from nengo_spa.pointer import SemanticPointer
 from nengo_spa.modules.product import Product as ProductModule
-from nengo_spa.exceptions import SpaNetworkError, SpaParseError, SpaTypeError
+from nengo_spa.exceptions import SpaConstructionError, SpaParseError, SpaTypeError
 
 
 class ConstructionContext(object):
@@ -1047,7 +1047,7 @@ class Action(Node):
 
     def construct(self, context):
         if context.bg is None or context.thalamus is None:
-            raise SpaNetworkError(
+            raise SpaConstructionError(
                 "Conditional actions require basal ganglia and thalamus.")
 
         # construct bg utility
--- a/nengo_spa/exceptions.py
+++ b/nengo_spa/exceptions.py
@@ -2,8 +2,36 @@ class SpaException(Exception):
     """A exception within the SPA subsystem."""
 
 
-class SpaNetworkError(SpaException, ValueError):
-    """An error in how SPA keeps track of networks."""
+class SpaConstructionError(SpaException):
+    """An error in the construction of SPA action rules."""
+
+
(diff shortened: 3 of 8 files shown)


Seanny123:
In `legacy-spa` the problem was doing something like:

`
model = spa.SPA()
model.blah = spa.State()
# do some other stuff...
model.blah = spa.State()
`

Re-assigning the attribute would break a lot of things, but quietly, so we made an error. Is this not longer a problem in this version?

jgosmann:
A lot of magic used to happen when assigning a SPA module, but nothing special (except setting the label which also is removed in this PR) is happening anymore. SPA modules are looked up via the assigned attributes when `Actions.build` is called. So either I don't know what I am overlooking or it is not a problem anymore.

jgosmann:
Added a commit to improve another error message and allow empty string arguments to `Vocabulary.populate`

Seanny123:
Any reason in particular why we wouldn't want to raise an error instead of doing nothing? I ask, because I haven't had to make this type of design decision before and thus don't know the thought behind it.

jgosmann:
Because right now it was raising an error in a case where I didn't want to raise it an error and had to insert an if-statement to handle it. To be more concrete: I was having a dynamically created list of names I wanted to have in the vocabulary that potentially can be empty and I was populating the vocabulary with `vocab.populate(';'.join(list_of_names))`. A join on an empty list produces the empty string and in this case nothing should be added to the vocabulary.

Note that `' ; '` will still give an error because this seems to be much more likely to be an actual coding error. But it is maybe not completely consistent. I could see an argument for having that not produce an error or keep the empty string as an error.

Maybe there should be some method to populate the vocab with a list of strings to make the `join` unnecessary and makes the different between `[]` and `['']` explicit. My reason for using the semicolon in a string syntax was that it for long lists it is way less annoying to type `'A; B; C` than `['A', 'B', 'C']` (maybe less of an issue on English keyboard layouts where no modifier key is needed to type `'`, but we should not expect everyone to use an English keyboard layout).

Seanny123:
I'm tempted to make that an error, but maybe bring it up at the next meeting to get more opinions on this?

Seanny123:
Alternatively, if you want to remove that one error check into another pull request, I'm fine with the rest of the changes.

jgosmann:
Sounds good, I dropped the commit and will merge on your approval.

@ikajic, feel free to still comment after a merge on the missing key exception for vocabularies in associative memories (#29). Any further improvements would likely build onto the changes in this PR and can be implemented in another PR if required.

Seanny123:
This LGTM now. Thanks for discussing this in depth with me.

jgosmann:
np, can you officially approve the PR?

PR #43, opened 2017-06-13
Fix transforms on first vector in vocab.

jgosmann:
**Motivation and context:**
For the first vector that was added to a vocabulary the transform would not be applied. For examples `vocab.populate('A.unitary()')` would not produce a unitary vector if no other vectors have been added before.

none

**How has this been tested?**
Added a regression test.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_spa/vocab.py
+++ b/nengo_spa/vocab.py
@@ -77,27 +77,28 @@ class Vocabulary(Mapping):
         candidate pointer with lowest maximum cosine with all existing
         pointers is returned.
         """
-        if len(self) == 0:
-            best_p = pointer.SemanticPointer(self.dimensions, rng=self.rng)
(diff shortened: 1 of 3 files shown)


nengo_spa/vocab.py
```diff
@@ -111,6 +112,8 @@ def __iter__(self):
 
     def __getitem__(self, key):
         """Return the semantic pointer with the requested name."""
+        if key == '__tracebackhide__':
```
Seanny123: Is this used only for testing? That's what I got from consulting StackOverflow. If it is, what test is it specifically used in? You may want to a comment, since it's an unusual piece of code that may merit explanation.

jgosmann:
Added the requested comment.

PR #61, opened 2017-06-30
Add output methods to IdentityEnsembleArray

jgosmann:
**Motivation and context:**
Adds `add_output` and `add_neuron_output` methods to `IdentityEnsembleArray` to provide the same API as `EnsembleArray`.

none

**How has this been tested?**
added new tests

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,14 @@ Release History
 0.2.1 (unreleased)
 ==================
 
+**Added**
+
+- Add ``add_output`` and ``add_neuron_output`` methods to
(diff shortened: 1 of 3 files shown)


Seanny123:
This wouldn't be used if someone wanted to make a learning connection from a SPA ensemble, right? In that case, they would use a `State` population? Or can you do learning from an ensemble array without problems in the case of SPA? Was this documented/discussed somewhere?

jgosmann:
In theory you could use `add_output` to create an output with a learning rule on that connection, but in practice I probably would not. The limitations are exactly the same as for the standard `EnsembleArray`.

Issue #70, opened 2017-08-27
Action rule complexity limited by recursion limit


jgosmann:
The current implementation of the `construct` function of the AST nodes uses recursion which limits the maximum complexity of action rules. This is probably not a problem for most actions rules, but maybe we should consider rewriting this to use a different approach (there are non-recursive tree-traversal algorithms using a stack separate from the call stack, also see the end of [this mind-bending talk](http://www.dabeaz.com/finalgenerator/)).

jgosmann:
Resolved with the current syntax that is constructing networks as early as possible.

Issue #54, opened 2017-06-26
Action rule syntax


jgosmann:
It was noted during the last lab meeting that the current action rule syntax might be misleading with its use of the equals sign. Currently, action rules use this syntax: `utility --> sink = source`. Please use this issue to suggest alternatives to the `-->` and `=` symbols.

jgosmann:
Give #63 I feel that the usage of the equals sign in more of a problem because the rest of the action rule language ends up being much closer to normal Python. To me a syntax like

* `sink <- source`
* `sink <-- source`
* `source -> sink`
* `source --> sink`

seems much better (potentially even allowing both directions of the arrow) as it highlights more the graph structure and direction of information transmition. However, that requires to change the syntax of the basal ganglia utility value (currently using `-->`) which is very similar (or identical) to this. So far I don't have very many ideas. Potentially a colon `:` could work, but is more easily overlooked: `utility: source -> sink`.

Tagging @astoeckel (who initially brought this issue up, I believe) and tagging @tcstewar who's opinion is always welcome in SPA questions.

tcstewar:
I really have no idea on this one...  :(  But I do like the idea of shaking up the syntax a bit.  The current `-->` syntax was mostly stolen from ACT-R, so there's no great reason to keep it....

I wonder if we could also bring in some indenting to help with this?  And also to help with complex action rules starting to become unreadable?  I'm thinking something like this:

Current approach:
```python
    actions = spa.Actions([
        'dot(vision, WRITE+SAY) --> verb=vision',
        'dot(vision, YES+NO+HI+BYE+OK) --> noun=vision',
        'dot(phrase, VERB*WRITE) - 2*dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK)'
            '--> hand=phrase*~NOUN',
        'dot(phrase, VERB*SAY) - 2*dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK)'
            '--> speech=phrase*~NOUN',
        'phrase = verb*VERB + noun*NOUN',
        ])
```

Alternate approach:
```python
    actions = spa.Actions('''
        dot(vision, WRITE + SAY):
            verb <- vision
        dot(vision, YES+NO+HI+BYE+OK):
            noun <- vision
        dot(phrase, VERB*WRITE) - 2*dot(vision, WRITE+SAY+YES+NO+HI+BYE+OK):
            hand <- phrase*~NOUN
[...]

PR #72, opened 2017-09-15
New action rule syntax

jgosmann:
**Motivation and context:**
This PR implements the new action rule syntax as discussed in #54.

It is another step towards implementing our own compiler. There is now a tokenization step, a parsing step to construct a parse tree and then that parse tree is converted to the abstract syntax tree (AST). The process of constructing actual Nengo objects from the AST has not changed (for the most part).

As the new syntax is very Python-like, we just use the Python tokenizer provided by the `tokenize` module.

For the parsing we define a context-free grammar with a set of more generally usable Python classes on the tokens provided by the tokenizer. However, we don't detail the complete grammar for evaluable expressions because later all those tokens are gathered and converted back into a string for evaluation. The end of an expression is detected by certain delimiter symbols is the grammar. The constructed grammar also provides the parser to create a parse tree from the token sequence. This parser is a simple recursive top-down parser. That implies certain limitations:
* recursion limit limits complexity of expressions (should not be a problem and makes only sense to resolve together with #70)
* (Simple) top-down parsers do not terminate for left-recursive grammars, but our grammar is not left-recursive (but one could (though should not) use the general classes for the grammar definition to define a left-recursive grammar)
* The parser implements no form of backtracking. For our given grammar this is fine, but it can make the order of alternatives in our grammar matter and might not be sufficient for future extensions.

The created parse tree is mostly putting the tokens into a hierarchy, but also adds minimal annotations to certain nodes in the tree.

In the next step, a visitor pattern is used to convert sets of nodes in the parse tree to AST nodes.

I added a new AST node `ActionSet` that groups all the actions of one basal ganglia. This puts the BG/thalamus construction into the AST (where everything else is constructing) which is more consistent. It also allows multiple `ifmax` blocks implemented in different basal ganglias.

none, except that this is a major API change that affects everything using action rules

**How has this been tested?**
existing adjusted tests and new tests

As this is a rather large PR, I recommend going through it commit by commit which builds the SPA compiler in a bottom up approach. Each commit leaves the code in a state where all tests pass.

1. The first commit just adds the `Peekable` generator wrapper (a comparably small commit). This class is not used at this stage, but will be later as the parsing of the action rules requires look-ahead in the sequence of tokens.
2. The next commit adds a wrapper for Python's `tokenize` module. There are some difference between the Python 2.7 and 3.x versions of this module, so this helps to add a layer to hide some of these differences.
3. Then we start getting to the core of things. 33756b85a249e5018265226ecda6b4918055b8d0 implements classes that allow to define a context-free grammar and in parallel create a parser for this grammar. Read the docstrings for more details (there are some limitations). The parser will operate on the sequence of tokens produced by the Python tokenizer.
4. The next step is to convert the parse tree (which is obtained when parsing with the grammar of the previous step) into the abstrac syntax tree (AST). Most of the AST code already existed, but this commit moves it to the `nengo_spa.compiler` package and splits the big module into three modules:
  * `ast_types.py` only defines the classes/objects for marking types in the AST.
  * `ast_nodes.py` defines the `Node` base class, the specific nodes, and some related helpers.
  * `ast.py` defines classes and functions for the conversion of the parse tree (and `ConstructionContext` which is used for constructing the actual Nengo objects).
5. The final step switches the `Actions` implementation over to use the new action syntax and updates the tests and examples.
6. All the remaining commits are to improve the error messages.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Python 2.7 compatibility
- [x] Update tests and examples
- [x] Tests for converting parse tree to AST
- [x] Documentation
- [x] Code cleanup
- [x] Cleanup commit history.
- [x] check for remaining FIXMEs and TODOs
- [x] check which exceptions are raised and whether SPA specific exceptions are still required


jgosmann:
The code base in this PR should be in a working state now (on Python 3, for Python 2.7 there is still some work required). This should others (@tcstewar, @Seanny123, @astoeckel) allow to play around with this branch while I still clean up things. It would especially useful, if you could see what sort of exceptions you get when writing invalid action rules (and what sort of exception and information you would expect in those situations).

jgosmann:
Python 2.7 compatibility is added (makes the code less nice in some spots ...)

I still have to cleanup the commit history and then see about exceptions/error messages. But feel free to test already.

jgosmann:
Ready for review!

nengo_spa/compiler/parsing.py
```diff
@@ -0,0 +1,319 @@
...
+            Peekable generator providing the upcoming tokens.
+
+        Returns
+        -------
+        bool
+        """
+        raise NotImplementedError()
+
+    def read(self, tokens):
+        """Reads the tokens excepted by the rule and returns derived objects.
```
Seanny123: Is this supposed to be "accepted" or "excepted"?

nengo_spa/compiler/tests/test_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+    assert rule.read(tokens) == [
+        MatchToken(type=tk.NAME, string='bar')]
+
+    assert not rule.accept(tokens)
+    with pytest.raises(parsing.RuleMismatchError):
+        rule.read(tokens)
+
+
+def test_either_chaining_gathers_rules_in_single_object():
+    rules = [parsing.Terminal() for i in range(3)]
```
Seanny123: I'd prefer if you used `for _ in range(3)` just to indicate that the iteration variable isn't used, but that's a nitpick I can fix myself.

nengo_spa/compiler/tests/test_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+    either = rules[0] | rules[1] | rules[2]
+    assert all(r1 is r2 for r1, r2 in zip(either.rules, rules))
```
Seanny123: Could you use `itertools.combinations` or something to distinguish this test from `test_chaining_gathers_rules_in_single_object`?

nengo_spa/compiler/tests/test_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+        MatchToken(type=tk.NAME, string='bar')]
+
+    assert not rule.accept(tokens)
+    with pytest.raises(parsing.RuleMismatchError):
+        rule.read(tokens)
+
+
+def test_chaining_gathers_rules_in_single_object():
+    rules = [parsing.Terminal() for i in range(3)]
+    either = rules[0] + rules[1] + rules[2]
```
Seanny123: Should this be called `chain` instead of `either` assuming this test is distinct from `test_either_chaining_gathers_rules_in_single_object`.

nengo_spa/compiler/tests/test_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+    parse = any_number.read(tokens)
+    expected = code.split(' ')
+    if expected == ['']:
+        expected = []
+    assert [x[1] for x in parse] == expected
+
+
+def test_group():
+    tokens = peekable_tokens('foo bar')
+    grammar = parsing.Group('grammar', parsing.Group(
```
Seanny123: Is `grammar` and `word` a significant identifier here or could you also just say `parent` and `child`?

nengo_spa/compiler/tests/test_parsing.py
```diff
@@ -0,0 +1,143 @@
...
+        'words', parsing.AnyNumber(parsing.Terminal(type=tk.NAME))))
+
+    assert grammar.accept(tokens)
```
Seanny123: Could you describe the block of this test with a comment? It's hard to understand how this is testing all the intended functionality.

nengo_spa/compiler/tests/test_ast.py
```diff
@@ -0,0 +1,436 @@
...
+            network.state = spa.State(d)
+
+    ast = AstBuilder().build_expr(bare_tokens('network.state'))
+    assert ast == [Module('network.state', network.state)]
+    assert str(ast[0]) == 'network.state'
+    ast[0].infer_types(model, None)
+    assert ast[0].type.vocab == network.state.vocabs[d]
+
+
+def test_scalar_multiplication():
```
Seanny123: I feel like this could be broken up into a separate tests. One that tests inferring vocabulary directly and one that tests inferring vocabulary from the network.

nengo_spa/compiler/parsing.py
jgosmann: I think it can be either. The rule defines certain tokens it expects which are the same tokens that need to be there for the rule to accept the tokens. However, the `accept` method only looks one (maybe two) tokens ahead and doesn't check all expected tokens.

nengo_spa/compiler/tests/test_parsing.py
jgosmann: I don't think so. `itertools.combinations` is doing something else than `zip`.

nengo_spa/compiler/ast.py
```diff
@@ -0,0 +1,288 @@
...
+        self.root_network = root_network
+        self.bg = bg
+        self.thalamus = thalamus
+        self.bias = None
+        self.sink = sink
+        if active_net is None:
+            active_net = root_network
+        self.active_net = active_net
+
+    def subcontext(self, bg=None, thalamus=None, sink=None, active_net=None):
```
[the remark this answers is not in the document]
jgosmann: This isn't new code in the PR, it has just been moved around. When Nengo objects are constructed from the AST, this happens in a certain context (active network, which BG and thalamus things are connected to etc.). This context can be changed by the parent node for its child node (e.g. to get nested networks). Usually only a few attributes of the context change in that case while the rest stays the same. This method makes it easier to create such a subcontext without passing all the arguments to the constructor.

nengo_spa/compiler/ast.py
```diff
@@ -0,0 +1,288 @@
...
+    def build_expr(self, parse_tree):
+        expr = self._untokenize_expr(parse_tree)
+        return [eval(expr, {}, self.scope)]
+
+    def build_sink(self, parse_tree):
+        expr = self._untokenize_expr(parse_tree)
+        obj = eval(expr, dict(self.scope.locals), dict(self.scope.globals))
+        return [Sink(expr.strip(), obj)]
+
+    def _untokenize_expr(self, parse_tree):
```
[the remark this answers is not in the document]
jgosmann: This method expects the parse tree of the `expr` rule in the grammar (which actually won't be a tree, but just a flat list of tokens). So `parse_tree` represents a single expression (or what the SPA grammar defines as such, it might differ from the exact definiton of a Python expression). The parse tree (i.e. the list of tokens) is untokenized, that is, it is converted back into a string which is then returned.

nengo_spa/compiler/tests/test_ast.py
jgosmann: These tests have just been moved to a new place. Which doesn't mean that the structure couldn't be improved, but if we split up this test, it should be done to all other tests too. Right now there is one test function corresponding to each type of AST node and they all have a similar structure.

jgosmann:
Replied to @Seanny123 comments and did the easy fixes.

nengo_spa/compiler/grammar.py
```diff
@@ -0,0 +1,166 @@
...
+
+    Does not accept potential leading encoding or newline tokens.
+    """
+
+    def accept(self, tokens):
+        return Terminal(type=tk.INDENT).accept(tokens) or action.accept(tokens)
+
+    def read(self, tokens):
+        match = []
+        if Terminal(type=tk.INDENT).accept(tokens):
```
Seanny123: Maybe explain the two cases?

nengo_spa/compiler/tests/test_ast.py
```diff
@@ -0,0 +1,436 @@
...
+from nengo_spa.compiler import tokens
+from nengo_spa.compiler.ast import AstBuilder
+from nengo_spa.compiler.ast_nodes import (
+    Action, ActionSet, ApproxInverse, DotProduct, Effect, Effects, Negative,
+    Module, Product, Scalar, Sink, Sum, Symbol)
+from nengo_spa.compiler.ast_types import (
+    TActionSet, TEffect, TEffects, TScalar, TVocabulary)
+
+
+def bare_tokens(expr):
```
Seanny123: Maybe explain why this is used for tests that don't need a full parse tree structure?

nengo_spa/compiler/ast_nodes.py
```diff
@@ -1146,4 +963,56 @@ def evaluate(self):
...
+        if len(self.effects.effects) > 0:
+            effect_str = "".join(
+                "\n    " + line for line in str(self.effects).split("\n"))
+        else:
+            effect_str = "\n    pass"
+        return 'ifmax {utility}{name}:{effects}\n'.format(
+            utility=self.condition, name=name_str, effects=effect_str)
+
+
+class ActionSet(Node):
```
Seanny123: Maybe explain why this object is required to create a single root node?

Seanny123: Also, how it helps handle multiple bg and thal?

nengo_spa/compiler/grammar.py
```diff
@@ -0,0 +1,172 @@
...
+
+        def __str__(self):
+            return "'->'"
+
+        def __repr__(self):
+            return "ConnectL2R()"
+
+    connect_l2r = ConnectL2R()
+else:
+    connect_l2r = Terminal(type=tk.OP, string='->')
```
astoeckel: Maybe add a comment here indicating that the "->" operator is present in later Python versions

nengo_spa/actions.py
```diff
@@ -1,287 +1,113 @@
...
-    @property
-    def bg_actions(self):
-        return [a for a in chain(self.actions, self.named_actions.values())
-                if isinstance(a, Action)]
+    try:
+        parse_tree = parse(actions)
+    except RuleMismatchError as err:
+        raise _rule_mismatch_err_to_syntax_err(actions, err)
+    except tokens.TokenError:
+        raise SyntaxError("unexpected EOF while parsing actions")
```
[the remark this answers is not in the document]
jgosmann: According to [the Python documentation](https://docs.python.org/3/library/tokenize.html#tokenize.TokenError) it is the only possible cause of this error. I will keep the name, though, as `tokens` is only intended to be a thin wrapper for `tokenize` to give a compatible interface in Python 2.7 and 3.

jgosmann:
@Seanny123 I addressed your comments and opened issue #75.

Seanny123:
Those changes are great! Some of the tests are changing, but once they pass I'll approve this.

PR #103, opened 2017-11-08
Allow noop actions with pass keyword.

### LINKED ISSUE #101: Noop rules
Is there a way to implement noop rules in the new syntax?  I'm thinking of something like the old syntax where you could do
```
a --> x=y
b --> 0
```
in order to implement the logic `if a > b: x=y`.  I'm not sure how exactly that worked under the hood, but the `0` essentially did nothing (since we don't want to do anything if `b > a`).

The equivalent
```
ifmax a:
    y->x
elifmax b:
    0
```
doesn't work, but is there some other way to accomplish this in the new syntax?

jgosmann:
**Motivation and context:**
Allows actions without any effect by using Python's `pass` keyword:
```python
spa.Actions('''
    ifmax 0.5:
        pass
''')
```

Fixes #101.

none

**How has this been tested?**
added a test

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -35,6 +35,10 @@ Release History
 - Allow the ``spa.Actions`` string to be empty.
   (`#107 <https://github.com/nengo/nengo_spa/issues/107>`_,
   `#109 <https://github.com/nengo/nengo_spa/pull/109>`_)
+- The ``pass`` keyword can now be used to create blocks in action rules that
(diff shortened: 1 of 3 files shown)


Seanny123:
Note that `flake8` is failing for some reason.

jgosmann:
Will fix before merging.

PR #108, opened 2017-11-09
Add object-based Action interface

drasmuss:
**Motivation and context:**
The recent efforts to add comments/syntax highlighting to the spa syntax made me think that it might be worth trying out an interface that relies only on normal python syntax, rather than the string parsing approach.  This is just a simple proof of concept to show what that might look like.  It uses all the same network construction logic, it just replaces the string parsing step with a `numpy-esque` function interface.

E.g., instead of
``` python
spa.Actions("""
    ifmax dot(a, b):
        c -> d
    elifmax dot(c, d):
        a * b -> c
""")
```
you would do
``` python
Actions([
    (dot(a, b), route(c, d)),
    (dot(c, d), route(bind(a, b), c))
])
```

Pros are that you get all the normal python stuff for free (syntax highlighting, comments, indentation/formatting/pep8, documentation via docstrings, autocomplete, etc.) and don't have to worry about writing/maintaining our own grammar/parser.  Main con I think is that it's a bit more verbose (but I'm not an advanced `nengo_spa` user, so there may be more advanced features of the string syntax that I haven't tapped into).

I'm not advocating that we get rid of the string-based approach, or even that we would merge this new approach into master.  Just putting it out there as something worth thinking about (I'll probably use it in my own work).

**How has this been tested?**
I copied `test_actions.py` and replaced all the `spa.Actions` with the new syntax (left the `spa.Actions` equivalent in as a comment, to show the difference).

However, I'm sure there are some cases I'm missing that aren't in `test_actions.py`.  This is definitely a WIP.

**Types of changes:**

**Checklist:**

- [ ] I have updated the documentation accordingly.

**Still to do:**
Documentation, more thorough testing, discussion


jgosmann:
I agree that it would probably be less error-prone on the implementation side.

But:
* While it gives us all the benefits of the Python syntax, it also limits us to that syntax (which might not be as expressive for the use case).
* The string syntax seems to me much more approchable for someone new to SPA as it is more self-explanatory
* I also feel the string syntax is visually much easier to parse quickly when glancing over (but that is subjective).

Though, those might be fixable ... most importantly I would want to add operator overloading so that one can write `a * b` instead of `bind(a, b)` etc.

drasmuss:
> Though, those might be fixable ... most importantly I would want to add operator overloading so that one can write a * b instead of bind(a, b) etc.

Yeah, I tried to figure out how to do that, but it's a bit awkward because you'd have to overload the operators on the spa Modules despite those operators only making sense in the context of an action.  E.g.,

```
a = spa.State()
b = spa.State()
c = spa.State()

Actions(route(a*b, c)) # this would be cool
a * b                  # this is also a valid statement now
```
So, not the end of the world, but it's a bit weird to define those operations outside the Actions context.  I was trying to think of some fancy thing that'd use the `Actions` as a context manager, and only define those operations within that context?  But I couldn't think of any clever ways to do that other than making the context manager go and monkeypatch `spa.Networks`, which seems bad.

jgosmann:
I'm not so much worried about `a * b` being a valid statement outside of the actions context. It would allow easier reusability like:
```python
bound = a * b
Actions(route(bound, c))
```
Though, there are also some questions: Would `a * b` suffice to add the binding network or does it need to be referenced in an Action rule? What happens if it gets referenced multiple times? How do we infere the right vocabularies if bare semantic pointers are used?

Seanny123:
To offer my own *very personal* opinion, the object-based syntax reminds me of [Polish Notation](https://en.wikipedia.org/wiki/Polish_notation) and LISP.  I've always had a hard time reading Polish Notation, because I see mathematics as a kind of flow. Consequently, reading `+ 3 4`, I have to mentally transpose `+` in between `3 4`.

Similarly, when I see `route(bind(a, b), c)`, I have to mentally transpose the route to imagine the neural "flow" that will result from this statement. I don't have to do exert as much mental effort for `a * b -> c`. 

But that's just my opinion.

Also, I'm tagging @tcstewar because this object notation reminds me of a less painful version of his initial attempt at defining a dictionary-like SPA syntax.

jgosmann:
> the object-based syntax reminds me of Polish Notation and LISP

I had the same associations ...


> his initial attempt at defining a dictionary-like SPA syntax.

Is this the legacy SPA syntax or even another kind of syntax?

Seanny123:
> Is this the legacy SPA syntax or even another kind of syntax?

This is ancient SPA syntax from Nengo 1.4.

drasmuss:
Just pushed a commit that adds operator overloading.  I started out by adding them to the `Network` class like we discussed, but that ended up with some tricky circular dependencies (Network imports ast_nodes which imports Network).  I'm sure those could be resolved, but it'd require some fairly significant reorganization of a bunch of code, whereas what I'm really aiming for with this thing (in its current form at least) is a minimal, standalone, drop-in replacement for `spa.Actions`.  

So I went back to the context manager monkey-patching idea which, in the end, didn't end up being as messy as I feared it might be (although it's definitely a bit hacky/magical).  So how it works now is you can do
```
a = State()
b = State()
with Actions() as actions:
    #a+b, a*b, -a, ~a, a-b are all defined within this context
    actions.add_block(route(a*b, c))

# a + b not defined here
```
I think encapsulating that functionality within the Actions context actually ends up being a good conceptual thing too.  As Jan pointed out, if you're just allowed to put `a*b` anywhere in your code, there are a lot of questions about how that should be interpreted (should I be able to do `nengo.Probe(a*b)`?).  This way it's clear that those operations are only defined wrt creating spa Actions, which (hopefully) clears up some of the ambiguity.

Seanny123:
Out of curiosity and ignorance, how hard would it be to implement something like:

```
a = State()
b = State()
with Actions() as actions:
    actions.add_block((a*b, route_to, c))
```

This *very personally* looks more understanding to me as I said before. I don't think you should implement this, I just want to know what options are available. I tried looking at the code, but couldn't quite see how much this would disrupt things. I think it definitely changes how things are parsed, given I switched a function to an argument?

jgosmann:
Potentially crazy idea, how about something like:
```python
import nengo_spa as spa
from nengo_spa import route, utility, dot

with spa.Actions():
    route(a * b).to(c)
    with utility(dot(state, 'A')):
        route('B').to(state)
    with utility(dot(state, 'B')):
        route('C').to(state)
```

Seanny123:
How would this case manage multiple basal ganglia? Would they each be mapped to a separate `spa.Actions` object?

jgosmann:
I would say, you'd create multiple `spa.Actions` objects. We could also do another level of nesting to create multiple BGs in one `Actions` object, but I think I like that less because it adds boilerplate indentation (is that a thing?) in many simple cases.

drasmuss:
> Out of curiosity and ignorance, how hard would it be to implement something like:

It'd definitely be doable (I'd probably go with something more like `add_block((a, "->", b))`.  I'd be kind of inclined against it; I think that while `(a, "->", b)` might be easier to parse for experienced `spa`/`nengo_spa` users, it'd be harder to parse for new users.  I think if you see `route(a, b)` and don't understand it, it's easy to figure out what's going on by going and looking up the documentation for the `route` function (a familiar workflow for anyone).  But if you see `(a, "->", b)` and don't know what's going on, how do you discover that?  Perhaps go to the documentation for the `add_block` function I guess, where we could document this behaviour.  Not the end of the world, but feels like a bit more of a leap for new users to me.  But that's not a strong opinion, I'd be ok with `(a, "->", b)` if that's what people prefer; as you say, this is all pretty subjective!

An other option might be to specify routing with a dict, like `add_block((dot(a, b), {c: d})`.  It's kind of like infix notation, and you could read it like "c maps to d".  I'd still probably prefer just `route(c, d)`, but it's another option.

drasmuss:
> route(a*b).to(c)

We could also do `route(a*b, to=c)` (gets the same semantics in there, and then people can still just do `route(a*b, c)` if they want.

xchoo:
Just spitballing. Isn't `>` an operator as well? We could potentially overload that as well, if people really are so inclined.

jgosmann:
*Very subjective* but `route(source).to(sink)` seems visually easier to pass (maybe because I'm trained on picking out strings enclosed in parenthesis?).

Also, given the choice, I would prefer to force people to write the `to` (which is possible in newer Python versions for keyword arguments, but we don't want to drop support for older Python yet).



> Just spitballing. Isn't > an operator as well? We could potentially overload that as well, if people really are so inclined.

Interesting ... but I think one should be very careful with overloading widely used operators to do non-standard things. That might obscure things more because it is not evident that `>` in that context is doing something else than comparison.

drasmuss:
Just summarizing ideas here:

Routing source (`a*b`) to sink (`c`):
``` python
route(a*b, c)
route(a*b, to=c)
route(a*b).to(c)
(a*b, "->", c)
a*b >> c
```

jgosmann:
I did not realize that Python has a `>>` operator, but that one might not be the worst choice as at least in C++ it has a history of being overloaded and isn't that commonly used in Python.

drasmuss:
Added two commits:

a07d92a

This automatically adds created actions to the active `Actions` context (as suggested by Jan).  So instead of
``` python
with Actions() as actions:
    actions.add_block(
        (dot(a, b), route(a*b, c)),
        (dot(b, c), route(a*b, d)))
```
you just do
``` python
with Actions():
    cond(dot(a, b), route(a*b, c))
    cond(dot(b, c), route(a*b, d))
```
At first this struck me as a big magical, but then I realized that it's basically the same as automatically adding Nengo objects to the Network context.  So we're already teaching users to expect this kind of behaviour, and it does eliminate a lot of boilerplate.

Note that different bg blocks are specified by different contexts, so
``` python
with Actions() as actions:
    # bg block
    ...
with actions:
    # these rules will be in a new bg
    ...
```

6434d4a

This overrides the `rshift` operator for routing, so instead of
``` python
with Actions():
    route(a*b, c)
```
you can do
``` python
with Actions():
    a*b >> c
```
I actually don't mind this too much, I think the `rshift` operator is uncommon enough that users won't have a lot of preconceptions that are violated.  I'm still fine with `route(a*b, c)`, but if everyone is keen on the infix operator then I think this is a good option.

jgosmann:
> Note that different bg blocks are specified by different contexts

This is surprising to me. I would expect everything to be part of the same basal ganglia if I reuse the same object, but not, when I create a second object. I'm not sure how easy it is to adjust the implementation. Alternatively, there might not even be a reason anymore for the user to explicitly instantiate the `Actions()` object. So we could have a module level instance and do:

```python
with actions:
    # first bg
with actions:
    # second bg
```

In that case it is not as evident that the same object instance is reused for the context and it looks more like a special keyword.

drasmuss:
> This is surprising to me. I would expect everything to be part of the same basal ganglia if I reuse the same object, but not, when I create a second object. I'm not sure how easy it is to adjust the implementation. 

Yeah that'd be an easy switch to make, no strong preference either way on my end.

jgosmann:
## Referring to semantic pointers

An important feature in the string syntax is that it treats names starting with capital letters as semantic pointers with saves a lot of typing for certain models (and especially in the examples). I think we cannot implement a custom lookup for global identifiers (and even if it would probably not be a good idea). So how do we provide access to semantic pointers without too much boilerplate. The currently supported solutions are:

* `vocab['Name']`
* `vocab.parse('Name')` (where this allows expressions like `2 * Name + A * B`).

### Both are problematic for a number of reasons:

* In simple models the vocabulary is often not explicitely defined which makes it awkward to get the `vocab` object.
* Prefixing every semantic pointer with `vocab` (or even `vocab.parse`) name adds a lot of boilerplate (the string syntax finds the correct vocabulary automatically).
* The symbol chain `['..']` is always a bit awkward to type (on my keyboard layout) and adds 4 characters to potentially very short names (I often just use names like V0, V1, V2, ...)

### Potential solutions:

1. Assign the pointers to variables: `sp_name = vocab['Name']`. This reduces the typing effort if the same pointer is used multiple times, but makes things worse if it is only used once. (This will also likely break type checking, i.e. making sure that pointers from different vocabularies are not combined when they should not be combined, because a `SemanticPointer` has no information about its vocabulary associated).
2. Add a helper function that assigns the semantic pointers from a vocabulary to the current `locals` dict. That might overwrite other variables and pollute the name space, but reduces boilerplate.
3. Have a one or two letter helper function to declare a string as referring to a semantic pointer: `sp('Name')`. Reduces boilerplate, but is still a bit awkward to type.
4. Use a special object with custom `__getattr__`: `sp.Name`. Easy to type, but somewhat magical or opaque.

## Types that (potentially) need support in the overloaded operators

Hopefully, this is a complete list:

* SPA modules: already existing in this PR, non-problematic as we have full control over these objects.
* `SemanticPointer` objects: like module unproblematic. But not sure if we should allow those as they currently do not now which vocabulary they are part of and allow accidentally mixing different vocabularies.
* Numpy arrays: I am leaning to disallow these because the `*` operator has a different meaning in SPA. Better to explicitly convert the array into a `SemanticPointer` and use that (if we allow it).
* Whatever type of object is used to solve the first part of this post.
* The integer (float?) `0` which depending on context can be the scalar 0 or the null vector/semantic pointer. We cannot overload the operators for built-in types. But I suppose it is fine to handle everything as a normal scalar Python operations until that scalar is combined with a SPA module at which point we can overload the operator appropriately. We could also enforce the usage of `nengo_spa.pointer.Zero` for the semantic pointer semantics (depending on how with handle `SemanticPointer` objects). Or we could only allow it to have the semantic pointer semantics in `vocab.parse` and equivalent expressions.
* The integer (float?) `1` which depending on context can be the scalar 1 or the identity vector. Basically the same questions as for `0`. However, `nengo_spa.pointer.Identity` is the specific identity vector for circular convolution. Not sure if we want to anticipate [other binding operations](https://github.com/ctn-waterloo/internal_tech_reports/blob/master/Comparison%20of%20binding%20operations.ipynb) that have different identity vectors?
* All other scalars should be fine to have handled by Python until combined with a SPA module at which point we can use an overloaded operator.

Note I presented these as distinct points, but some of those connections are ultimately connected.

drasmuss:
> An important feature in the string syntax is that it treats names starting with capital letters as semantic pointers with saves a lot of typing for certain models (and especially in the examples).

If I'm understanding you properly, that all works fine in the current object implementation.  E.g., the following are all valid and parsed correctly:

``` python
    with spa.Network() as model:
        a = spa.State(16)
        b = spa.State(16)
        with oact.Actions():
            "A" >> a
            "A" * a >> b
            "A * B" >> b
```
As long as one of the elements in the operation is either a) a spa Network or b) an ast Node, then the other elements of the operation will be cast to ast Nodes as normal.  And I don't think it is possible to write a valid spa effect that doesn't contain either a Network or a Node, so that should cover all cases (I think).

> The integer (float?) 0 which depending on context can be the scalar 0 or the null vector/semantic pointer.

I'd save `"0"` for the semantic pointer and leave the int/float `0` as a scalar.

jgosmann:
> As long as one of the elements in the operation is either a) a spa Network or b) an ast Node, then the other elements of the operation will be cast to ast Nodes as normal.

This adds cognitive load because one has to think about which parts need to be a single string and which parts must not be part of a string. For example, `'A * B' >> a` works, but `'A' * 'B' >> a` does not. Opposed to that `'A * b' >> a` does not work, but `'A' * b >> a` does work. To me it seems quite complicated to explain this to a novice user.

-----

btw: The `>>` operator has a lower precedence than the normal arithmetic operators, i.e. `a * b >> c` will be parsed as `(a * b) >> c` which is exactly what we want. :)

jgosmann:
Also a potentially dangerous case (in the sense of doing something unintended without warning or error): `'2 * A'` vs. `2 * 'A' = 'AA'`.

drasmuss:
> This adds cognitive load because one has to think about which parts need to be a single string and which parts must not be part of a string.

Yes, it's not perfect.  But as you point out, there are also downsides to forcing the user to explicitly cast all strings to symbols.  So the question is which option do we prefer: making users think about the difference between 
``` python
"A" * 2
vs 
"A * 2"
```
or making users think about the difference between
``` python
"A" * 2
vs
sym("A") * 2
```

Both seem roughly equivalently in terms of teaching/cognitive load, so I'm inclined towards the former since it's more parsimonious.  But I'm fine with either approach.

jgosmann:
In both `"A" * 2` and `sym("A") * 2` the user has to be aware that `A` refers to a semantic pointer and needs special notation (as opposed to `A * 2`).

jgosmann:
So I just got mixed up that `"A" * 2` is *not* `sym("A") * 2` or `"A * 2"` which is a strong indication to me that this confusion should be avoided. I think it also made me misunderstand your point slightly. So to hopefully address your actual point more directly:
* With `sym("A")` we can give the user the rule: Whenever you want to use a name referring to a semantic pointer in a vocabulary, use `sym("A")`. Everything else works as expected (hopefully).
* With pure strings the rule gets much more complicated: Whenever you want to use a name referring to a semantic pointer in a vocabulary, use string notation [this is equivalent to the other rule, specifying the syntax]. If you do calculations with constants on that pointer, all of those expressions go also into the string. If you are combining it in some way with a module, that operation does not go into the string.

drasmuss:
Updates:

e50896d

Since the decision in the dev meeting was to go ahead with the object syntax, I did some refactoring to get rid of the string API and replace it with the object actions.  The main change is that `ast_nodes.py` was merged with `object_actions.py` and renamed to `actions.py`, and most of the rest is just deleting code that is no longer being used.

b8647dd

Rename `cond` to `ifmax` as discussed in dev meeting.

9a7cea9

Add `sym("A")` to explicitly cast strings to symbols.  Disables symbol evaling (so `"A * 0.5"` is just a string).  We could leave the symbol evaling in, but it seemed confusing to have two different ways of manipulating symbols like that.

6cbf58a / 859de69

These are just updating unit tests/examples to the new syntax.  I haven't done the slow examples yet.

tcstewar:
Hey all, sorry I've been out of contact while this was developed.  I definitely like that this is being explored! It's definitely taking a different direction from when I explored a non-parser-based approach in nengo 1.4.... I'll take a closer look over the next few days, but I just wanted to note that I kinda like `sym.A` and I'm not sure it's any more magical than `sym("A")` (but much easier to type).  And `sym.A*2` doesn't lead to the confusion about `sym("A"*2)`......

xchoo:
Just wondering, is it possible (or even a good idea) to treat all strings in the action syntax as a (singular, without SPA operations) semantic pointer? i.e.
```
'a' * 'b' >> mem
```
is equivalent to:
```
sym('a') * sym('b') >> mem
```

And yes, I know that in python `'a' * 2 = 'aa'`, but I don't think it would be too much of a stretch that users can understand that every string in the `action` context is treated as a semantic pointer.

Also, with this, we can remove the restriction that all semantic pointers have to start with uppercases, no?

jgosmann:
Unfortunately no, because we cannot redefine the operators for built-in types (which probably wouldn't be a good idea in the first place). So `'a' * 'b'` will give you an error.

> Also, with this, we can remove the restriction that all semantic pointers have to start with uppercases, no?

I think so ...

xchoo:
Bummer. I don't suppose we can add new encoding methods to string literals?
i.e. `p'A'` being a semantic pointer (akin to `b'A'` meaning encode string in binary)?

drasmuss:
> Also, with this, we can remove the restriction that all semantic pointers have to start with uppercases, no?

Yeah, that's no longer a requirement on the Actions side, but I left it in because I wasn't sure if there was anything in the `Vocabulary` code that was based on that assumption.

jgosmann:
There might be a check for capital letters in there, but nothing that really requires that.

jgosmann:
I haven't looked at the code in detail yet, but at the moment we are monkey patching the `Network` class. There is another approach that I want to try out and might be cleaner:

1. Create a mixin class that defines the operators.
2. Have `nengo_spa.Network` derive from it (in addition to `nengo.Network`)
3. When defining inputs/outputs to SPA modules [dynamically add that mixin to the Nengo object](https://stackoverflow.com/questions/8544983/dynamically-mixin-a-base-class-to-an-instance-in-python).

With that it would be possible to allow “cortical” rules (not gated) outside of the `Actions` context; an idea maybe worth considering. Especially with some more thoughts I had about how the rules should work explained in the following.

Given some modules, we can construct the required networks and connections immediatly when the operator is applied. This gives shorthands to build certain networks and reuse parts, e.g.:

```python
state1 = spa.State(d)
state2 = spa.State(d)
state3 = spa.State(d)
state4 = spa.State(d)
bound = state1 * state2
unbound = bound * ~state1
bound + state3 >> state4
p = nengo.Probe(unbound.output)
```

It also gives a natural way of accessing constructed networks.

If such a module is combined with a fixed pointer, that boils down to a connection with a specific transform. This connection cannot be created without a target. So we could just create a `nengo.Node` as a target, but this could clutter a model quickly with “useless” nodes. Instead, I probably would just create an object that stores this information until a place to connect to is given (by `>>` or use in some operation with another module). This sort of object would be similar to the `Artifact` object used in the AST. Combination of purely fixed pointers and scalar could be handled in a similar way.

~~Potential downside: If `unbound = bound * ~sym.CUE`, you cannot to `nengo.Probe(unbound.output)` and the difference might not be clear.~~ To allow things like `unbound = bound * ~sym.CUE` and then `nengo.Probe(unbound.output)`, we could construct a `nengo.Node` with appropriate input connection on demand when the `output` attribute is accessed.

jgosmann:
I forgot to mention another reason to construct the required networks on operator application: It gives us the actual line in the traceback with doing a bunch of introspection of stack frames.

jgosmann:
I found a situation that is not covered by the current implementation:
```console
In [5]: with spa.Network() as model:
   ...:     a = spa.State(64)
   ...:     b = spa.State(64)
   ...:     with spa.Actions():
   ...:         a.output >> b.input
   ...:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-5-868131e07858> in <module>()
      3     b = spa.State(64)
      4     with spa.Actions():
----> 5         a.output >> b.input
      6

TypeError: unsupported operand type(s) for >>: 'Node' and 'Node'
```

I will take a look at fixing this in the mixin based implementation, so it might not be necessary to fix this here. Note that this is not as trivial as it might seem because `a.output` and `b.input` are normal Nengo nodes that do not know about the containing network that specifies the associated vocabularies which are required for type inference and type checking.

arvoelke:
> Since the decision in the dev meeting was to go ahead with the object syntax

Does anyone happen to remember what the main argument was to have this syntax over something like `spa.build(a * b >> c)` or `(a * b >> c).build()`, where the side-effects of building the connections and binding networks would be deferred to the explicit call?  Is anyone aware of any existing Python library that has behaviour in the side-effects of overloaded operators? I've been looking for some precedent, but not even SymPy appears to do anything like this (contrary to today's lab discussion).

jgosmann:
See #114. Main arguments are complexity and this:
> There are also some things which are not possible with #108 at the moment. Most importantly module inputs and outputs (e.g., bind.input_a) are not supported. That could be fixed, but would require to go through all Nengo objects and monkey patch them (and unpatch them afterwards).

Also, not mentioned in #114: Getting in error pointing to the actual line that is causing the error.

arvoelke:
Whoops, I intended the above comment for #114 -- not this implementation. To clarify, I was asking about the difference between the current implementation `(a * b >> c)`, and something like `(a * b >> c).build()`. You could still do all of the error-checking and validation in the same place that it's currently done. But, instead of having the operators immediately build the Nengo objects, these side-effects would be deferred until the call to `build`.

jgosmann:
This still requires more complexity. Also, having to type `(...).build()` for every line seems annoying.

PR #114, opened 2017-12-11
Mixin ops

jgosmann:
**Motivation and context:**
This PR implements a new Nengo SPA action rule syntax that uses operator overloading and can thus be directly written as Python code without the encapsulation into a string. While this allows us less flexibility, it simplifies the code and makes it less error prone as we no longer need to implement our own parser.

Furthermore, this PR changes the AST implementation to create new networks as early as possible. This avoids building a complete syntax tree and simplifies things even further as it requires less classes to represent different things in the syntax tree and the code paths for building the AST and constructing Nengo objects from it can be unified. However there is one downside: when using a Semantic Pointer symbol, the corresponding vocabulary is unknown. Previously, the vocabulary was inferred from the complete expression. This is not possible without constructing the complete syntax tree. The code still tries to infer the vocabulary from the second operand in operation, but does not use information from higher up in the syntax tree. While this, might be a slight limitation in some cases, it should rarely matter in practice because the specific vocabulary is usually only required in operations that involve a SPA module that already has a specific vocabulary (though there are some other cases). Also, the vocabulary could easily be specified with `sym.reinterpret(vocab)`.

A new feature is that you can use `SemanticPointer` instances in action rules now. This was required to implement the `translate` operation on semantic pointer symbols, but seems reasonable independent of that as it means that the action rule syntax just works on pretty much all SPA objects. Supporting this required to track to which vocabulary a `SemanticPointer` belongs, but this might be useful anyways as it introduces more type safety into the semantic pointer arithmetic and the associated vocabulary is easily overwritten with `reinterpret`.

This is what the new syntax looks like:

```python
vocab = spa.Vocabulary(64)
vocab.populate('A; B')

with spa.Network() as model:
    state = spa.State(vocab)
    x = spa.State(vocab)
    y = spa.State(vocab)

    spa.sym.A >> state  # transmit semantic pointer symbol 'A' to state
    vocab.parse('A * B') >> state  # this was not possible before
    spa.sym.A * vocab['B'] >> state  # also works
    spa.sym.A * spa.sym.B >> state  # or this
    x * y >> state  # of course modules can be used in action rules
    (x + spa.sym.A) * y - vocab['B'] >> state  # and more complex stuff mixing things

    # and this is how action selections works:
    with spa.ActionSelection() as act_sel:
        ifmax(dot(state, spa.sym.A), spa.sym.B >> state)
        ifmax(dot(state, x * ~y), x + y >> state, spa.sym.A >> x, spa.sym.B >> y)
        default_utility = ifmax(0, spa.sym.A >> state)  # default action, we manually connect a utility input
    nengo.Connection(nengo.Node(lambda t: t), default_utility)  # manual utility input
    act_sel.bg   # basal ganglia
    act_sel.thalamus  # thalamus
```

Another thing that now is possible is to assign partial rules and reuse them multiple times:

```python
with spa.Network() as model:
    bound = item * pos
    0.7 * bound + 0.7 * old >> updated
    bound * ~spa.sym.Pos1 >> first_item
```

Note that the first time a vocabulary is inferred it will be remembered to avoid the accidental usage of an assigned partial rule in the context of different vocabularies:

```python
v1 = spa.Vocabulary(32)
v2 = spa.Vocabulary(64)
with spa.Network() as model:
    a = spa.State(v1)
    b = spa.State(v2)

    some_pointer = spa.sym.Foo
    some_pointer >> a  # will infer the vocab for some_pointer as v1
    some_pointer >> b  # will raise an exception
```

This PR is based on #108, but mostly for the tests which have been updated to the new syntax for the most part in there. Much of the other code has been written ground-up.

**How has this been tested?**

Added new tests and updated existing tests.

It is probably best to look at the end state instead of going commit by commit because many there are commits setting things up for future changes and without the big picture the purpose of the changes might not be obvious. There might also be some commits where things went into one direction first and were reverted later to go into a different direction.

There is a certain hierarchy to the code and I recommend to follow it in reviewing the code (make sure to also look at the comments below the list).

1. Review `pointer.py` and associated tests. Operations on semantic pointers are closed, i.e. each operation will create another semantic pointer instance. Thus, you should not need any knowledge about other code at this point.
2. Take a look into `nengo_spa.ast.base`. It's short and just provides the base classes for what is to come.
3. Continue with `nengo_spa.ast.symbolic`. The `PointerSymbol` class handles operations with symbols for semantic pointer (where we cannot know the actual vector before the vocabulary has been inferred). Most operations on `PointerSymbol` are also closed, but some accept a `FixedScalar` too. Don't worry about how a number like `42` gets converted to `FixedScalar(42)` yet. There is one operation, `translate` which produces a `SemanticPointer` instance, but you reviewed this one earlier.
4. Now we get to `nengo_spa.ast.dynamic`, the core of things where we consider dynamic values that need to be implemented with actual networks. Some of the operator implementations become a bit more complicated because besides `DynamicNodes`, `PointerSymbol` and `SemanticPointer` are also allowed in the operations. To represent summed and transformed dynamic values `Summed` and `Transformed` are used. For the output of a module `ModuleOutput` is used.
5. Next you should take a look to the changes in `network.py`. Here a mixin class is defined that provides the operators with the necessary casting to the AST classes. Also, this mixin class is added dynamically to module inputs and outputs. This file will also reference the `__rshift__` operator, but to get the full picture of it, you have to look at the next step.
6. `actions.py` implements the `__rshift__` operator and action selection.
7. Finally, a small file `operators.py` defines an alternate way to call custom “operators” like `a.dot(b)` in a more function style way: `dot(a, b)`. Numpy supports both variants for many functions to and it is somewhat equivalent to an operator in Python where you have `a + b` (equivalent to the function style) that gets converted to the method call `a.__add__(b)`. Those functions like `dot(a, b)` support also any class that defines a `dot` method.

For the AST node(-like) classes there are two important methods: `connect_to` constructs all required network elements (if any) and connects them to a given target object. This function exists to avoid intermediate pass-through nodes (for example `Summed` would require them because the sum can only be created in the network by connecting all inputs to the same target). There exists also a `construct` method that constructs all required network elements without creating a connection from them. For fixed nodes this usually means creating just a node (`connect_to` creates the node and connects it). For dynamic nodes the network is usually already created and will be just returned, but in some cases (like `Summed`) an extra node that can be returned needs to be created. Thus, `connect_to` should be preferred as it more conservative in creating extra nodes, but `construct` is helpful to get access to the output of some SPA rule that is not routed to somewhere. For example two way to construct a circular convolution network:

```python
with spa.Network() as model:
    a = spa.State(d)
    b = spa.State(d)

    binding1 = spa.Bind(d)
    nengo.Connection(a.output, binding1.input_a)
    nengo.Connection(b.output, binding2.input_b)
    p1 = nengo.Probe(binding1.output)

    p2 = nengo.Probe((a * b).construct())  # connections implicit
```

Note that technically the network will be constructed before `construct` is called (there might be a better method name?), but the method call gives you access to the output node whereas without the call you only have access to the AST node. Maybe there should be a way to get the actual network too? Initially, I was thinking that any attribute access for a non-existing attribute on an AST node will be delegated to the network, but now I'm thinking that it might be to magical. Maybe some accessor method or attribute would be better?

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Test `FixedPointer` and `SemanticPointer` with `DynamicNode` and `PointerSymbol`.
- [x] Implement `SemanticPointer.transform_to`
- [x] Implement operators for `SemanticPointers` and make them `vocab` aware.
- [x] Code cleanup and documentation.
- [x] Better error messages.
- [x] Avoid introducing unnecessary nodes and (especially!) synapses.
- [x] Construct networks only once
- [x] Implement `@` operator.
- [ ] Automatic network construction on access of network attributes


drasmuss:
I just changed the base to make it easier to see what has changed vs #108

jgosmann:
This is now in a state where all the core functionality should be there. That means people could try it out. The next steps will be to clean up and document the code to get it ready for review.

jgosmann:
I think this is ready for review now. There is one more thing on the to do list, but I'd like some input on that from other people first. The AppVeyor tests are failing because of one test with too tight bounds, so nothing too critical. The Travis tests are failing because conda is currently completely broken on Travis and there is not much I can do about it.

Seanny123:
To be clear, what's the main difference between this and #108? Should I look at these PRs as being two competing implementations or does this implementation enable things not possible in #108?

jgosmann:
Yes, they are two competing implementations. #108 builds the full syntax tree and delays the construction of actual Nengo objects until the `Actions` context is left. While that allows for better type inference in some (hopefully) rare cases, it makes the construction of Nengo objects more complicated because the network construction context needs to be taken care of while it is implicit in this PR as networks are immediately constructed. Another difference is that #108 monkey patches the operators, while here the operators are always present.

There are also some things which are not possible with #108 at the moment. Most importantly module inputs and outputs (e.g., `bind.input_a`) are not supported. That could be fixed, but would require to go through all Nengo objects and monkey patch them (and unpatch them afterwards). The other thing that comes to my mind is the possibility to use `SemanticPointer` instances in the action rules (not possible in #108 at the moment).

Seanny123:
@jgosmann should I just hold off until @drasmuss reviews this? It seems pretty hard for someone like me to evaluate?

For instance, I would appreciate being pointed at where the monkey-patching in #108 occurs. Additionally, what does monkey-patching all the Nengo objects (and unpatching) them imply? Also, I would like a specific example of where type inference is superior in #108.

Only answer the above questions if you think it's useful to communicate their answers to everyone. Alternatively, I'm fine with waiting for the discussion between @drasmuss and you to progress.

drasmuss:
I probably won't have a chance to look at this for a while unfortunately (I'm not using `nengo_spa` in my project anymore), so I'd say go ahead with whatever approach you prefer.

jgosmann:
> It seems pretty hard for someone like me to evaluate?

Well, I shouldn't be the only one who understands that code (I might not be around forever to maintain it). If you cannot understand it, I might have to improve it. Also, didn't you review the parser and original AST? I think this PR might be less complicated, though you might have to read up a little bit on operator overloading in Python.


> For instance, I would appreciate being pointed at where the monkey-patching in #108 occurs.

[Right here.](https://github.com/nengo/nengo_spa/pull/108/files#diff-9c325e20d265997cd085f18b04d984e2R117)

> Additionally, what does monkey-patching all the Nengo objects (and unpatching) them imply?

Go through the whole model and for figure out for each object whether it is an module input/output and monkey-patch the operators in, then later undo the monkey patching in the same way.

> Also, I would like a specific example of where type inference is superior in #108.

I think there are only few cases where this is practically relevant. The only thing I can come up with right now is:
```python
v1 = spa.Vocabulary(64)
v2 = spa.Vocabulary(64)
with spa.Network() as model:
    state = spa.State(v1)
    out = spa.State(v2)
    state.reinterpret() * spa.sym.A >> out
```
Looking at the last line in whole, it can be inferred that `state.reinterpret()` and `spa.sym.A` should be interpreted in the context of `out`'s vocabulary `v2`. With this PR the vocabulary needs to be determined when `state.reinterpret() * spa.sym.A` is evaluated (because `spa.sym.A` needs to be resolved), but there is not way to known the correct vocabulary at that point. The proper way to do this with this PR would be `state.reinterpret(v2) * spa.sym.A`, i.e. being explicit what vocabulary should be used.

nengo_spa/ast/symbolic.py
```diff
@@ -0,0 +1,148 @@
...
+        if not isinstance(other, PointerSymbol):
+            return NotImplemented
+        type_ = infer_types(self, other)
+        return PointerSymbol(self.expr + '+' + other.expr, type_)
+
+    def __radd__(self, other):
+        return self + other
+
+    def __sub__(self, other):
+        other = as_symbolic_node(other)
```
Seanny123: The code block:
```
other = as_symbolic_node(other)
if not isinstance(other, SOMETHING):
     return NotImplemented
type_ = infer_types(self, other)
```

Seems to be repeated a lot, both in `symbolic.py` and in `pointer.py`. Would it be worth refactoring into a function?

nengo_spa/ast/symbolic.py
```diff
@@ -0,0 +1,148 @@
...
+        if not isinstance(other, PointerSymbol):
+            return NotImplemented
+        type_ = infer_types(self, other)
+        return PointerSymbol(self.expr + '-' + other.expr, type_)
+
+    def __rsub__(self, other):
+        return (-self) + other
+
+    def __mul__(self, other):
```
Seanny123: Given `__mul__` and `__rmul__` are identical, is it worth making a function `def mul(self, other)` and calling it in both cases?

Seanny123:
I noticed during testing that there should be a better error if you use `ifmax` outside of the `spa.ActionSelection` context.

Seanny123:
There should probably also be a better error if you input and invalid effect in `ifmax`. Like `ifmax(condition, 0 >> target, target)`.

Seanny123:
Finally, if you're routing to an invalid `sink`, such as `state_a >> ens` there should also be a better error. Currently what I got is:

```
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 53, in as_sink
    vocab = input_vocab_registry[input_]
  File "C:\Users\mr_bo\Anaconda3\lib\weakref.py", line 394, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x000002872CFD03B8; to 'Ensemble' at 0x0000028729A15470>

During handling of the above exception, another exception occurred:

  File "count.py", line 154, in <module>
    fin.comp_tot_fin >> thresh_ens
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 94, in __rshift__
    return as_ast_node(self) >> as_sink(other)
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 55, in as_sink
    raise SpaTypeError("Invalid sink.")
nengo_spa.exceptions.SpaTypeError: Invalid sink.
```

The error for invalid sources is even weirder:
```
  File "count.py", line 155, in <module>
    thresh_ens >> fin.tot_fin_simi
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 97, in __rrshift__
    return as_ast_node(other) >> as_sink(self)
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 40, in as_ast_node
    vocab = output_vocab_registry[output]
  File "C:\Users\mr_bo\Anaconda3\lib\weakref.py", line 394, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x000001AAA21403B8; to 'Ensemble' at 0x000001AA9EB824E0>
```

I'll make minimal reproducible examples for all of these sometime in the next week, but I gotta sleep now.

nengo_spa/ast/symbolic.py
jgosmann: `self` and `other` are switched, making them non-identical. This does not matter with circular convolution, but other binding operators might not be commutative.

jgosmann:
> Finally, if you're routing to an invalid sink, such as state_a >> ens there should also be a better error.

How exactly should the error be improved? Currently it is saying “Invalid sink.“, what other information would be useful?

Seanny123:
> Currently it is saying “Invalid sink.“, what other information would be useful?

The reason why the sink is invalid would be helpful. Something like `Invalid sink of type BLAH. Expected BLAH.` but I understand if all the expected types aren't listable.

Also, the error message I get on Windows is:
```
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 53, in as_sink
    vocab = input_vocab_registry[input_]
  File "C:\Users\mr_bo\Anaconda3\lib\weakref.py", line 394, in __getitem__
    return self.data[ref(key)]
KeyError: <weakref at 0x0000020B37DB0728; to 'Ensemble' at 0x0000020B37D997B8>

During handling of the above exception, another exception occurred:

  File ".\sink_err.py", line 11, in <module>
    state >> ens
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 94, in __rshift__
    return as_ast_node(self) >> as_sink(other)
  File "c:\users\mr_bo\git\mixin-ops\nengo_spa\network.py", line 55, in as_sink
    raise SpaTypeError("Invalid sink.")
nengo_spa.exceptions.SpaTypeError: Invalid sink.
```

Is there a way to avoid throwing two errors?

Seanny123:
As promised, here is the minimum code required to reproduce the errors I was talking about yesterday.

`ifmax` outside of `spa.ActionSelection` error:
```
import nengo_spa as spa
from nengo_spa import ifmax, dot

dims = 32

with spa.Network() as model:
    state = spa.State(dims)
    x = spa.State(dims)
    y = spa.State(dims)

    ifmax(dot(x, y), y >> state)
```

`ifmax` syntax error 1:
```
import nengo_spa as spa
from nengo_spa import ifmax

dims = 32

with spa.Network() as model:
    state = spa.State(dims)
    x = spa.State(dims)
    y = spa.State(dims)

    with spa.ActionSelection() as act_sel:
        ifmax(x, y >> state)
```

`ifmax` syntax error 2:
```
import nengo_spa as spa
from nengo_spa import ifmax, dot

dims = 32

with spa.Network() as model:
    state = spa.State(dims)
    x = spa.State(dims)
    y = spa.State(dims)

    with spa.ActionSelection() as act_sel:
        ifmax(dot(x, y), y)
```


Sink error:
```
import nengo_spa as spa
import nengo

with spa.Network() as model:
    state = spa.State(32)
    ens = nengo.Ensemble(100, 1)

    state >> ens
```

Source error:
```
import nengo_spa as spa
import nengo

with spa.Network() as model:
    state = spa.State(32)
    ens = nengo.Ensemble(100, 1)

    ens >> state
```

jgosmann:
> Is there a way to avoid throwing two errors?

Yes, but in Python 2 it requires rather ugly code.

Seanny123:
> Yes, but in Python 2 it requires rather ugly code.

In that case, don't bother and maybe make a not of it on the Nengo forum as another reason to drop Python 2 support?

nengo_spa/ast/symbolic.py
jgosmann: I added a commit to do such a refactoring as much as possible and reasonable. There are two things that prevent a further reduction of the duplicated code:

1. The type `SOMETHING` is sometimes the class that is just being defined. Thus, we cannot define a decorator because it needed to be instantiated with a reference to that class when it is not yet fully defined. We also cannot use a normal function because in the case that the type does not match the outer function must return `NotImplemented`.
2. The last line `type_ = ...` could be refactored out, but then the value of `type_` needs to be passed in as a function argument which changes the function signature of the operators to a non-default signature. That seems confusing to me and the code duplication seem to be the lesser evil.

jgosmann:
> Is there a way to avoid throwing two errors?
> 
> Yes, but in Python 2 it requires rather ugly code.

Wasn't that ugly after all ...

jgosmann:
@tcstewar Do you agree?

jgosmann:
ping @tcstewar Any comments on the new syntax?

jgosmann:
> The only thing that was slightly confusing to me was that translate now requires a target vocabulary, rather than inferring it from context. I'm okay with that if it was intentional or an implementation requirement, but I didn't see anyone mention that change so I thought I'd point it out. It'd be nice if the simpler version could be done, but it's not a big deal.

This is an implementation requirement. When the `translate` function call gets evaluated the context is not known (the result could even be used in multiple contexts).

> The only other thing that I've noticed is that it'd be nice to tidy up the Network layout (in terms of keeping things in subnetworks and labelling them), so that it looks somewhat reasonable in nengo_gui

That might not be possible. Also, a network might be reused in different places, so it is not clear into which network is should belong. One thing that might be possible is to shove everything defined in an `ActionSelection` context into one network.

On the user side, things could easily be grouped, by explicitely defining them in appropriate network contexts.

jgosmann:
> I did find one tiny edge case that doesn't work:

that is fixed now

jgosmann:
I rebased this to master and added three commits (869f952d7cd4b36ad8353c24f650b8c0b5b99208, e38c700a280b7d89b0f0d2ca5055befabe8b2832, 86727d708f9b54e77cc956e67a07c69a2413850f) to (hopefully) make all CI tests pass. This includes bumping of Numpy version numbers, the minimum required version is now 1.8.

jgosmann:
And one more commit ...

Issue #126, opened 2018-02-07
Improve naming of networks


jgosmann:

Seanny123:
Close via #130

PR #130, opened 2018-02-16
Improve network names and grouping

jgosmann:
**Motivation and context:**
This is an attempt to improve the naming and grouping of modules; mainly for a better display in the Nengo GUI. 

### Naming
* All Nengo SPA modules get a default label. This label, unfortunately, takes precedence even when the module is assigned to a variable in Nengo GUI, i.e. `name = spa.State(64)` will display the name 'State' in Nengo GUI, not 'name'. The obvious workaround is `name = spa.State(64, label='name')`.
* Symbolic expression will display as such under the generated nodes (this was already the case prior to this PR). Example: `spa.sym.A * spa.sym.B * 0.3' will display as 'A*B*0.3'.
* Non-symbolic Semantic Pointer nodes will be label with 'Semantic Pointer (XXd)` where XX is the dimensionality.
* The utility nodes for basal ganglia inputs (the `ifmax` return value) are currently unlabeled. The user will not really interact with these as long as they are not assigned to a variable. So this make sense imo.

### Grouping

* Everything in the context of an action selection object gets now grouped into a network. I'm not sure whether that's better or worse. Unfortunately, this also includes all connections, so these won't be displayed in Nengo GUI until the network is 'opened'.
* Everything else gets shoved into the network where it is defined (as before). This allows the user to easily group things manually into separate networks. It, however, puts the connections again into the same network which can hide them in Nengo GUI.

Improvements on this grouping might hard because everything gets constructed immediately. For example if an `a*b` gets constructed, it is not known if it is part of `a*b >> state1` or `(a*b) - c * d >> state2`. Thus, there isn't really a way to make decisions based on the context without changing ownership of network components after the fact. One also has to keep in mind, that expressions can be used in multiple places, e.g. `tmp = a * b; tmp >> state1; tmp - c * d >> state2`.

It is also not necessarily clear to me how things should be grouped optimally if we had complete freedom. It might change depending on context. 

Addresses #126 and #127.

Let me know what you think of these things. Requested a review from @bjkomer too as he was intending to use the new SPA syntax himself.

none

**How has this been tested?**
Looked at it in Nengo GUI.

**Types of changes:**

**Checklist:**

**Still to do:**
Further improvements? Discussion?

### DIFF
--- a/nengo_spa/actions.py
+++ b/nengo_spa/actions.py
@@ -139,7 +139,9 @@ class ActionSelection(Mapping):
     active = None
 
     def __init__(self):
+        super(ActionSelection, self).__init__()
         self.built = False
+        self.bias = None
         self.bg = None
         self.thalamus = None
         self._utilities = []
@@ -158,11 +160,12 @@ class ActionSelection(Mapping):
 
     def __exit__(self, exc_type, exc_value, traceback):
         ActionSelection.active = None
+        if exc_type is not None:
+            return
+        self._build()
 
+    def _build(self):
         try:
-            if exc_type is not None:
-                return
-
             if len(RoutedConnection._free_floating) > 0:
                 raise SpaActionSelectionError(
                     "All actions in an action selection context must be part "
@@ -173,6 +176,7 @@ class ActionSelection(Mapping):
         if len(self._utilities) <= 0:
             return
 
+        self.bias = nengo.Node(1., label="bias")
         self.bg = dynamic.BasalGangliaRealization(len(self._utilities))
         self.thalamus = dynamic.ThalamusRealization(len(self._utilities))
         self.thalamus.connect_bg(self.bg)
@@ -186,13 +190,11 @@ class ActionSelection(Mapping):
                     self.thalamus.connect_fixed(
                         index, effect.sink.input, transform=effect.transform())
                 else:
-                    self.thalamus.construct_gate(
-                        index, net=nengo.Network.context[-1])
+                    self.thalamus.construct_gate(index, self.bias)
                     channel = self.thalamus.construct_channel(
                         effect.sink.input, effect.type)
                     effect.connect_to(channel.input)
                     self.thalamus.connect_gate(index, channel)
-
         self.built = True
 
     def __getitem__(self, key):
--- a/nengo_spa/ast/base.py
+++ b/nengo_spa/ast/base.py
@@ -54,6 +54,16 @@ class Node(object):
         raise NotImplementedError()
 
 
+class Noop(Node):
+    """Node that has no effect."""
+
+    def connect_to(self, sink, **kwargs):
+        pass
+
+    def construct(self):
+        raise NotImplementedError("Noop nodes cannot be constructed.")
+
+
 class Fixed(Node):
     """Base class for AST nodes that provide a fixed value."""
 
--- a/nengo_spa/ast/tests/test_dynamic.py
+++ b/nengo_spa/ast/tests/test_dynamic.py
@@ -45,7 +45,8 @@ def test_binary_operation_on_modules(Simulator, op, suffix, rng):
         sim.run(0.5)
 
     assert sp_close(
-        sim.trange(), sim.data[p], vocab.parse('A' + op + 'B'), skip=0.3)
+        sim.trange(), sim.data[p], vocab.parse('A' + op + 'B'), skip=0.3,
+        atol=0.3)
 
 
 @pytest.mark.parametrize('op', ['+', '-', '*'])
--- a/nengo_spa/modules/assoc_mem.py
+++ b/nengo_spa/modules/assoc_mem.py
@@ -60,8 +60,8 @@ class AssociativeMemory(Network):
 
     def __init__(
             self, selection_net, input_vocab, output_vocab=None, mapping=None,
-            n_neurons=50, label=None, seed=None, add_to_container=None,
-            vocabs=None, **selection_net_args):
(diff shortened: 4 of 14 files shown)


Seanny123:
Nice! I have a few comments.

In the case where inputs to the network are known, would it be useful to add that information into the label? For example, in the code:

```
import nengo
import nengo_spa as spa

dim = 32
vocab = spa.Vocabulary(dimensions=dim)

words = ['ORANGE', 'APRICOT', 'CHERRY', 'STRAWBERRY', 'APPLE']

for word in words:
    vocab.populate(word)



with spa.Network('AssociativeMemory', seed=1) as model:
    a = spa.State(vocab, label='a')
    b = spa.State(vocab, label='b')
    
    model.assoc_mem = spa.ThresholdingAssocMem(threshold=0.3, input_vocab=vocab)
    a*b >> model.assoc_mem

```

Would it be helpful if the `Bind` network was named `Bind(a,b)`?

jgosmann:
> Would it be helpful if the Bind network was named Bind(a,b)?

Maybe? But it is also redundant information, might produce long names, and complicates the implementation.

Seanny123:
> redundant information, might produce long names, and complicates the implementation.

Legit. I was thinking about all the interim populations I had created in other models and how I managed them. However, I think I should try building large networks in the new syntax and see how they display before commenting further.

jgosmann:
It might be worth to get this merged before the summer school?

Seanny123:
To review this properly, I have to test it with the Nengo GUI. However, I'm a bit confused which version I should test against. Should I test against the last release or the Master? Is there going to be a release before the Summer School? I realise making a new release is being discussed in a [Nengo GUI issue](https://github.com/nengo/nengo-gui/issues/951).

jgosmann:
I think any “recent” release or development version of the GUI should be fine to test that. The networks should be displayed the same. If you want to have the SPA related plots, you will need nengo/nengo-gui#891.

Seanny123:
This is definitely a huge improvement from before. However, as you mention, I think grouping all the connections inside of the Action Selection object, meaning it just looks kind of floating, is a bit confusing. I think we've talked about this before and you do seem to discuss it in the original PR text, but would you mind explaining in greater depth the different options? Maybe it's fine for the summer school and we can revisit how to visualize bg/thal connections later after the Nengo GUI refactor?

I wasn't really able to test the symbolic nodes. Could you give me some code showing them off so I could experiment with the feature?

jgosmann:
> I think grouping all the connections inside of the Action Selection object, meaning it just looks kind of floating, is a bit confusing. I think we've talked about this before and you do seem to discuss it in the original PR text, but would you mind explaining in greater depth the different options? Maybe it's fine for the summer school and we can revisit how to visualize bg/thal connections later after the Nengo GUI refactor?

I suppose one of the first question to ask (and I'm sure we discussed this before, but can't remember the details) whether the GUI should hide connections that are defined inside a network, but connect to things outside. If the answer is that the GUI should handle this differently this might become a non-issue for Nengo SPA, otherwise there are a few things to think about:

* Technically the connections are created within the ActionSelection network, so technically they are handled correctly. But what is technically correct might still be unintuitive or surprising to the user.
* It is not clear how to avoid the connections being part of the ActionSelection network:
  * We could delay the construction until the ActionSelection context is left, but this adds complexity as we need to keep track of the connections to add and might be surprising as everything else gets constructed immediately.
  * We could change the ownership of the connections after leaving the action selection context. But that might be surprising again.
  * We could potentially create the connections within a different context when they are created. Seems like the least surprising, but potentially still surprising.
  * ...?

In all of these cases I am not sure how the correct network to add things to could be determined (I'd need to look at the code in detail). And what network should they be added to? The parent of the ActionSelection network? (We don't have parent references in the networks ...) The network one above in the context stack? (Which tends to be the parent network, but might not always be the parent network.)

There is definitely some careful thinking required and to me it is not clear what the optimal solution would be even ignoring the implementation details.

Seanny123:
I think part of the problem with connections is they communicate two different things to me:

1. Anatomical connections. Which networks are physically connected to each other.
2. Functional connections. Where the data flows in a network.

When it comes to the bg+thal, I want the connections to do both at the same time and I feel like that's not a good idea.

All the things you mentioned as "surprising" violate the first principle, but to me, the current state violates the second. Unsurprisingly, [we've discussed this before](https://github.com/nengo/nengo-gui/issues/922).

After talking to @jgosmann, we've decided:
- Forcing people to wrap the action selection results in a network to make a pretty layout is less baffling than hiding the connections completely.
- Later, we will be able to pursue more ideal cases such as:
  1. Forcing the action selection network to be open by default
  2. Having multiple views in the Nengo GUI to support large scale networks where you can differentiate the anatomical connections from the functional ones

@tcstewar do you have any opinion on this?

jgosmann:
@tcstewar Would you like to comment? It would be nice to get out another release with the changes in this PR before the summer school (and I am probably camping :tent: :evergreen_tree:  on the weekend, so it needs to happen tomorrow).

tcstewar:
> Forcing people to wrap the action selection results in a network to make a pretty layout is less baffling than hiding the connections completely.

That makes sense to me.  I like that reasoning.

> Later, we will be able to pursue more ideal cases

One possibility would be to adjust nengo_gui such that it looks inside closed Networks (but only one layer deep) and draws Connections it finds there.  That would solve this nicely, I think....

jgosmann:
I squashed the last commit and ammended it to remove the `hash` method that didn't have any purpose with the recent change. Also added another test, to meet the codecov criterion.

Still looks good?

PR #151, opened 2018-05-17
Fix manifest

jgosmann:
**Motivation and context:**
The `global-exclude *.tox*` in the `MANIFEST.in` did not work and my `.tox` directory was stilll included in the distribution. `prune *.tox` does work.

Also adds the URL to the documentation to `setup.py`.

none

**How has this been tested?**
`python setup.py sdist bdist_wheel`

**Types of changes:**

**Checklist:**

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -14,10 +14,10 @@ graft docs
 prune docs/_build
 prune dist
 prune examples
+prune *.tox
 
 # Patterns to exclude from any directory
 global-exclude *.git*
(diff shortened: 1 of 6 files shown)


jgosmann:
(Not sure if you noticed, but code spell did catch those typos, but only on this PR and not the earlier ones.)

jgosmann:
check-manifest only seems to catch files missing from the distribution, but not files that should not be in there.

Issue #139, opened 2018-03-23
Add support for scalars changing over time


Seanny123:
## Problem

`spa.Transcode` won't allow for scalar inputs, so there's no way to input a time-varying scalar input with the `spa.Actions` syntax.

For example, if I make a custom module with a scalar input:
```
class GatedMemory(spa.Network):

    vocab = VocabularyOrDimParam('vocab', default=None, readonly=True)
    neurons_per_dimension = IntParam(
        'neurons_per_dimension', default=200, low=1, readonly=True)
    
    def __init__(self, vocab=Default, neurons_per_dimension=Default, **kwargs):
        super(GatedMemory, self).__init__(**kwargs)
        
        self.vocab = vocab
        self.neurons_per_dimension = neurons_per_dimension
    
        with self:
            self.mem = workingmemory.InputGatedMemory(
                self.neurons_per_dimension, self.vocab.dimensions)
        
        self.input = self.mem.input
        self.gate = self.mem.gate
        self.reset = self.mem.reset
        self.output = self.mem.output
        
        self.declare_input(self.input, self.vocab)
        self.declare_input(self.gate, None)
        self.declare_input(self.reset, None)
        self.declare_output(self.output, self.vocab)
```

I'm going to have a problem routing something to `gate` or `reset`, other than a constant input, using the `spa.Actions` syntax.

```
dimensions = 32

model = spa.Network()

with model:
    model.spa_in = spa.Transcode("OKAY", output_vocab=dimensions)
    model.g_mem = GatedMemory(dimensions)
    
    model.spa_in >> model.g_mem
    0 >> model.g_mem.gate
```

## Possible Solution: New Module

Make a module called `spa.ScalarNode(func)` which allows for time-varying inputs and outputs.

## Possible Solution: Modify `spa.Transcode`

Allow to specify the `output_vocab` or `input_vocab` as `1` thus indicating returning a scalar? This feels a little bit like over-loading a concept.

## Possible Solution: Allow any Nengo object with a 1D output to be used

Somehow?

Seanny123:
@tcstewar do you have any opinions on this?

jgosmann:
I'm in favour of allowing any Nengo object with 1D output.

Seanny123:
How exactly would you accomplish this? I ask, because I can't tell if this is something I can edit myself or not.

jgosmann:
[...]

PR #157, opened 2018-05-18
Allow scalar Nengo objects in action rules

jgosmann:
**Motivation and context:**
Allowing standard Nengo objects for scalars in action rules simplifies writing these rules, especially when routed by the action selection system. (Otherwise connections and gates have to be created manually.)

none

**How has this been tested?**
added a test

**Types of changes:**

**Checklist:**

### DIFF
--- a/docs/examples/custom_module.ipynb
+++ b/docs/examples/custom_module.ipynb
@@ -15,7 +15,7 @@
    "name": "python",
    "nbconvert_exporter": "python",
    "pygments_lexer": "ipython3",
-   "version": "3.6.1"
+   "version": "3.6.5"
   },
   "name": ""
  },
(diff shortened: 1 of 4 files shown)


jgosmann:
Not sure if there is any place in the documentation where this feature should be mentioned. I don't think we are explaining the syntax that detailed anywhere ...

Seanny123:
The one place the syntax comes up is the custom module example, which is what caused me to open #139 in the first place. Consequently, I updated the example to reflect this new capability, but I don't think it needs to be highlighted anywhere else, other than a changelog entry.

docs/examples/custom_module.ipynb
```diff
@@ -122,6 +122,8 @@
       "    model.config[GatedMemory].neurons_per_dimension = 150\n",
       "    \n",
       "    spa_in = spa.Transcode(\"OKAY\", output_vocab=dimensions)\n",
+      "    gate_in = nengo.Node(lambda t: 1 if t < 0.1 else 0)\n",
```
jgosmann: Is this used anywhere?

Seanny123: Oops. Now it is.

PR #162, opened 2018-05-24
Add support for hierarchies of vocabularies

jgosmann:
**Motivation and context:**
Sometimes it can be useful to construct larger vocabularies from smaller ones. This PR adds according functionality. Adding this functionality turned out to require more careful thought than I expected and adds quite a bit of complexity. So there can be an argument made to not merge this feature.

This PR adds two functions `combine_vocabs` and `pair_vocabs`. `combine_vocabs` creates a new vocabulary that contains all the keys of the vocabularies passed in. This requires that there are no duplicate names. By default it checks for the max. similarity constraint to avoid surprises, but this can be deactivated (because the only way to ensure it is to chose sufficient dimensionalities and hope that the god of randomness is on your side ;) ). 'pair_vocabs' creates a new vocabulary with all the bound pairs. Duplicate names are allowed (because new keys will be constructed). Again a check for the max. similarity is performed that  can be disabled. The keys of the paired vocabulary are the concatenation of the keys of the input vocabularies with a some symbol, by default `_`. So if there are SPs `A` and `B` in vocabs 1 and 2, the paired key would be `A_B`. Note that we cannot use `*` because it is not allowed in a SP name. I'm not sure whether I am happy with the name `pair_vocabs`, so other suggestions are welcome. But I would like to avoid anything `circular convolution` as other binding operations might be supported in the future.

Because these functions create new vocabularies, updates to the original vocabularies are not reflected. I think this is less surprising than if they were reflected and makes it easier to add further pointers to any vocabulary but exclude them from the combine or pair operation.

Technically, the creation of new vocabularies also means that they are all independent and thus incompatible types. This turns out to be rather awkward and requiring a lot of `reinterpret` calls in the action syntax. Thus, an attempt is made to keep track of the relations of vocabularies (and this is were the main complexities come into play). Each vocabulary tracks it supersets (i.e. other vocabularies that contain the same set of keys and potential other keys). Adding pointers to a subset will not destroy the superset relation. I think this is reasonable especially consdering the larger picture of type compatibility. Two vocabularies that are in a subset/superset relation are compatible with each other in either direction. Thus a connection can be made from a module with the subset vocabulary to a module with the superset vocabulary or the other way around (just allowing one direction would require a larger number of changes given how the type system works).

For the pairing of vocabularies, each vocabulary stores a list of factors which are the IDs (we don't want to keep a reference to allow garbage collection) of vocabularies that would generate this paired vocabulary. There might be multiple such combinations. So far writing `a * b` was not possible if `a` and `b` had different vocabularies, but now we want this to result in the paired vocabulary and allow the routing to a module that uses such a vocabulary. However, the code doing the work to implement `a * b` does not have access to the place where `a * b` gets routed and cannot obtain the desired vocab instance. Thus, a `LazyPairedVocab` instance is created which implements the `Vocabulary` interface (or at least some part of it that is important). This allows to access the paired pointers with keys like `A*B`. Note that the actual multiplication operator is used here for technical reasons. It is used to split the key into the subkeys to dynamically calculate the requested vector (with `A_B` it would be unclear if this needs to be split into two subkeys or is a single key in one of the paired vocabularies). The reason for creating theses lazily evaluated vocabularies is that the pairings of vocabularies can makes the vocabulary size grow exponentially and thus we should avoid creating too many of these intermediary vocabs. The `LazyPairedVocab` also provides the factors and it can be checked whether another vocabulary was (or can be) created from the same factors, in which case the vocabularies are compatible and the routing is allowed without explicit `reinterpret`.

Note that when `a` and `b` use the same vocabulary the result will also use the same vocabulary (instead of the vocab paired with itself). If either `a` or `b` has no specific vocabulary assigned, it will be inferred from the other operand and then the previous rule is applied (use the same vocab for the result).

I'm not super happy with the name `LazyPairedVocab`, so other ideas are welcome. I'm also unsure whether to include it in the documenation. Users should usually not instantiate it, but they might end up getting an instance of this class if doing stuff like `x = a * b; x.vocab`, and because it is not a full vocabulary implementation it might behave in surprising ways until converted to a proper vocabulary with `to_vocabulary()`.

none

**How has this been tested?**
added tests

tbd

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Documentation
- [x] Add support for two different vocabs to `Bind`.
- [x] Python 2 compatibility
- [x] Test `create_subset` semantics for superset tracking

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -23,6 +23,15 @@ Release History
 0.5.1 (unreleased)
 ==================
 
+**Added**
+
+- Methods ``join_vocabs`` any ``pair_vocabs`` for construction vocabularies
+  from smaller vocabularies. Furthermore, these vocabularies and subset
+  vocabularies will keep track of their hierarchical relationships to reduce
+  the number of required ``reinterpret`` operations.
+  (`#79 <https://github.com/nengo/nengo_spa/issues/79>`_,
+  `#162 <https://github.com/nengo/nengo_spa/pull/162>`_)
+
 
 0.5.0 (June 1, 2018)
 ====================
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -41,9 +41,6 @@ Release History
   similarity could not be obtained.
   (`#117 <https://github.com/nengo/nengo_spa/issues/117>`_,
   `#158 <https://github.com/nengo/nengo_spa/pull/158>`_)
-- Possibility to name Vocabulary instances for debugging.
-  (`#163 <https://github.com/nengo/nengo_spa/issues/163>`_,
-  `#165 <https://github.com/nengo/nengo_spa/pull/165>`_)
 
 **Changed**
 
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -14,4 +14,5 @@ of how ``nengo_spa`` works.
    examples/spa_sequence
    examples/spa_sequence_routed
    examples/spa_parser
+   examples/combining_vocabularies
    examples/vocabulary_casting
--- a/docs/examples/combining_vocabularies.ipynb
+++ b/docs/examples/combining_vocabularies.ipynb
@@ -1,0 +1,315 @@
+{
+ "metadata": {
+  "kernelspec": {
+   "display_name": "nengodev",
+   "language": "python",
+   "name": "nengodev"
+  },
+  "language_info": {
+   "codemirror_mode": {
+    "name": "ipython",
+    "version": 3
+   },
+   "file_extension": ".py",
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython3",
+   "version": "3.6.1"
+  },
+  "name": ""
+ },
+ "nbformat": 3,
+ "nbformat_minor": 0,
+ "worksheets": [
+  {
+   "cells": [
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "# Combining vocabularies\n",
+      "\n",
+      "Sometimes a model will have distinct vocabularies, for example one for shapes and one for colors, but in certain places you might need to combine these vocabularies into larger vocabularies. This examples demonstrates different use cases and possibilities to do so."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
+     "input": [
+      "%matplotlib inline\n",
+      "\n",
+      "import nengo\n",
+      "import nengo_spa as spa"
+     ],
+     "language": "python",
+     "metadata": {},
+     "outputs": [],
+     "prompt_number": 1
+    },
+    {
+     "cell_type": "markdown",
+     "metadata": {},
+     "source": [
+      "## Subsets\n",
+      "\n",
+      "It is possible to create a subset from a vocabulary containing only specific keys. This can be helpful for excluding some Semantic Pointers from a cleanup (associative) memory."
+     ]
+    },
+    {
+     "cell_type": "code",
+     "collapsed": false,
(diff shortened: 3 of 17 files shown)


Seanny123:
FYI, this is breaking due to Python 2, not having `list.clear()`.

jgosmann:
The main implementational work for this PR is done and updated the description. I would not mind some feedback while still working on the remaining points.

CHANGES.rst
```diff
@@ -37,6 +37,9 @@ Release History
   similarity could not be obtained.
   (`#117 <https://github.com/nengo/nengo_spa/issues/117>`_,
   `#158 <https://github.com/nengo/nengo_spa/pull/158>`_)
+- Possibility to name Vocabulary instances for debugging.
+  (`#163 <https://github.com/nengo/nengo_spa/issues/163>`_,
```
Seanny123: I think you meant to add this change to the other PR?

Seanny123:
This seems fine to me, however it's hard for me to comment since I've never desired this feature. Do you have any examples of models that needed this feature?

jgosmann:
There have been instances were I wanted this, but you can work around it, so I don't exactly recall them. But one example would be you have a vocabulary for shapes and one for colors that you might want to keep separate, but you also want to bind shapes with colors. With this PR it is easy to create all the desired vocabs (of course you could do it manually and use a bunch of `reinterpret`s). I also noticed that doing a cleanup to bound vectors isn't exactly easy to do with this PR. Not sure what the best solution would be without the PR, but with you can simply to this:
```python
v_shapes = spa.Vocabulary(...)
v_colors = spa.Vocabulary(...)

with spa.Network() as model:
    shape = spa.State(v_shape)
    color = spa.State(v_color)
    clean = spa.ThresholdingAssocMem(0.3, spa.pair_vocabs((v_shape, v_color)))
    shape * color >> clean
```

Seanny123:
Ah, that does seem like a good idea! Your approach does seem legitimate.

jgosmann:
I added an example notebook that might illustrate more use cases.

Seanny123:
Whelp, I need this now with Niel's giant model. I'll test this and get back to you.

jgosmann:
Why the hell is this failing with Python 2?

jgosmann:
I think I figured it out.

Seanny123:
FYI, for Jelmer's giant model, I mostly just created subsets of vocabularies and didn't end up needing the combination feature.

Also, the static check for this PR is failing due to a cyclic import.

jgosmann:
> Also, the static check for this PR is failing due to a cyclic import.

I know, you don't always have to tell me. ;)

jgosmann:
Eliminating the cyclic reference is non-trivial ...

jgosmann:
I am having second thoughts about this PR. It seems that in the end @Seanny123 was still required to use a lot of `reinterpret`s (#183), but the complexity of combining and especially pairing vocabs was not required. In addition the cyclic dependency cannot be eliminated (I believe) and might be an indication that the code is too complex or the architecture not well thought out.

Similar to #176 I feel that this PR tries to provide more convenience at the cost of purity. Or at least that I haven't thought out the theoretical foundations of what this PR means for the type system well enough. (What makes the SPA type system weird compared to usual type systems is that we work with types that are sets of semantic pointers and sets can be subsets of other sets and circular convolution technically does a Cartesian product of these sets, but often we want to stay in the original vocab.)

jgosmann:
As mentioned before this does not seem well enough thought out to me at the moment. I will close this PR soon if no one stops me.

jgosmann:
Going to close this now. Further reasons why this PR might not be a good idea are given [in this comment in another thread](https://github.com/nengo/nengo-spa/issues/183#issuecomment-404299257).

PR #185, opened 2018-06-14
Allow non-SPA networks as root network.

### LINKED ISSUE #174: Unhelpful error when using nengo.Network instead of spa.Network
When accidentally using `nengo.Network` instead of `spa.Network` it is possible to get an error that does not really point at the problem.

Try this code:
```python
import nengo
import nengo_spa as spa

model = nengo.Network()
with model:
    state = spa.State(16)
    state2 = spa.State(16)
    state >> state2
```

The error will be:
```
  File "/Volumes/Home/blubb/Library/Python/3.6/lib/python/site-packages/nengo_gui/page.py", line 220, in execute
    exec(compiled, code_locals)
  File "<nengo_gui_compiled>", line 8, in <module>
  File "/Volumes/Home/blubb/Library/Python/3.6/lib/python/site-packages/nengo_spa/network.py", line 117, in __rshift__
    return as_ast_node(self) >> as_sink(other)
  File "/Volumes/Home/blubb/Library/Python/3.6/lib/python/site-packages/nengo_spa/actions.py", line 38, in __rrshift__
    infer_types(self, other)
  File "/Volumes/Home/blubb/Library/Python/3.6/lib/python/site-packages/nengo_spa/ast/base.py", line 14, in infer_types
    type_ = coerce_types(*[n.type for n in nodes])
  File "/Volumes/Home/blubb/Library/Python/3.6/lib/python/site-packages/nengo_spa/types.py", line 157, in coerce_types
    raise SpaTypeError(reason + ": " + ", ".join(str(t) for t in types))
nengo_spa.exceptions.SpaTypeError: Different vocabularies: TVocabulary<16-dimensional vocab at 0x10dedf780>, TVocabulary<16-dimensional vocab at 0x10dedf2b0>
```

jgosmann:
**Motivation and context:**
Closes #174.

Instead of providing an error when accidentally using nengo.Network
and not a spa.Network, this makes it possible to use a nengo.Network.
The main reason the spa.Network was required, was to track the default
vocabularies. This commit adds a dictionary to that allows to store
the default vocabularies associated with root networks.

This is easier than figuring out whether we shoult raise an Exception
(because spa.Network is allowed to be used in nengo.Network by design).

none

**How has this been tested?**
added a test

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -20,6 +20,17 @@ Release History
    - Fixed
 
 
+0.5.2 (unreleased)
+==================
+
+**Fixed**
+
+- SPA modules will use the same default vocabularies even if not instantiated
+  in the context of a `spa.Network`.
+  (`#174 <https://github.com/nengo/nengo-spa/issues/174>`_,
+  `#185 <https://github.com/nengo/nengo-spa/pull/185>`_)
+
+
 0.5.1 (June 7, 2018)
 ====================
 
--- a/nengo_spa/network.py
+++ b/nengo_spa/network.py
@@ -1,4 +1,5 @@
 import inspect
+import weakref
 
 import nengo
 from nengo.config import Config, SupportDefaultsMixin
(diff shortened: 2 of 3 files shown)


Seanny123:
I guess I'm fine with this? I still feel like we should someone guide users to use `spa.Network` but I can't think of any way to do that and the alternative of giving a vague error is worse.

PR #188, opened 2018-06-29
Exclude keywords and unicode from SP names.

### LINKED ISSUE #187: vocab.parse handles some reserved Python keywords incorrectly
`vocab.parse('None')` gives `SpaParseError: The result of parsing 'None' is not a SemanticPointer.` instead of returning the Semantic Pointer named `None` or giving the correct error (if non-existent in strict vocab): `SpaParseError: Error parsing expression 'None' with 16-dimensional vocab at 0x7fc67d306e10: name 'None' is not defined`.

`vocab.parse('True')` and `vocab.parse('False')` will give the identity and zero vector respectively instead of the appropriate Semantic Pointer or error.

Note that this is somewhat related to #176.

jgosmann:
**Motivation and context:**
Because we rely on the Python interpreter for parsing we cannot use Python keywords as Semantic Pointer name. For some reason the test for unicode characters in Semantic Pointer names was commented out even though it shouldn't. Uncommented it and fixed the regex to require a match of the complete name.

Fixes #187.

none

**How has this been tested?**
existing and added tests

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -29,6 +29,10 @@ Release History
   in the context of a `spa.Network`.
   (`#174 <https://github.com/nengo/nengo-spa/issues/174>`_,
   `#185 <https://github.com/nengo/nengo-spa/pull/185>`_)
+- Disallow Python keywords, including None, True, and False, as well as unicode
+  characters from Python names.
+  (`#188 <https://github.com/nengo/nengo_spa/pull/188>`_,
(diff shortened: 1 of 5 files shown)


jgosmann:
Force pushed a few more changes to fix the tests.

jgosmann:
Another commit because Python 2 is annoying. :unamused:

PR #198, opened 2018-07-13
Add support for different binding operations

### LINKED ISSUE #69: Support for different binding methods
Circular convolution is only one possible choice for binding and [binding operations might have some advantages](https://github.com/ctn-waterloo/internal_tech_reports/blob/master/Comparison%20of%20binding%20operations.ipynb).

- We should consider to make easy to switch out the circular convolution default.
- We should consider implementing vector-derived transformation binding (in the linked document called vector-derived projection binding, but I came to the conclusion that “transformation” is more correct and will change it in the document with the next update)

jgosmann:
**Motivation and context:**
This adds the possibility to change the binding (and superposition) operation used by Nengo SPA. It also implements vector-derived transformation binding (VTB) as one alternate binding method.

The specific choice of a superpositiion operation and a binding operation is called an algebra in this PR (as it roughly fits the mathematical definition of an algebra or algebraic structure). Such in algebra is implemented by a class providing pure math implementations of the operations, methods to obtain special elements like the identity, and two methods to provide neural implementations of the two operators.

Each vocabulary (and `SemanticPointer` which might not have a vocabulary) is assigned such an algbra determining the operations used when operands get combined in the SPA syntax (it defaults to `CircularConvolutionAlgebra`). Because operands are required to have the same vocabulary, it is always clear what algebra to use, i.e., there is no situation in which the operands might have different algebras (except for `SemanticPointer` which raises an exception in this case).

Currently there is not way to change the default algebra, but one has to explicitely create a vocabulary with the desired algbra. Not sure whether there should be a way to set this default ...

Closes #69.

none

**How has this been tested?**

Tests have been added, updated, etc.
An `algebra` fixture has been added that will parametrize a test to run with both implemented algebras.

The commits should be self-contained, so that one could go commit by commit. But I revised some things multiple times and the back and forth might not be that helpful. Such it might be better to look at the final diff. There are multiple parts that stand mostly on its own:

* Added matrix multiplication network `nengo_spa/networks/matrix_multiplication.py`.
* Added module `nengo_spa/modules/superposition.py`.
* The interface for algebras `nengo_spa/algebras/base.py`.
* The circular convolution algebra `nengo_spa/algebras/cconv.py`. (Probably best to look at this one first as it uses familiar operations.)
* The VTB algebra `nengo_spa/algebras/vtb.py` with the binding network `nengo_spa/networks/vtb.py`.
* Several smaller changes in `nengo_spa/vocab.py` and `nengo_spa/pointer.py` and some other files to enable the use of algebras.
* All of this is accompanied with changes in the respective test files and documentation.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Extract the VTB network into its own class.
- [x] Try to cut down on the increased test duration.

### DIFF
--- a/nengo_spa/algebras/base.py
+++ b/nengo_spa/algebras/base.py
@@ -1,0 +1,231 @@
+class AbstractAlgebra(object):
+    """Abstract base class for algebras.
+
+    Custom algebras can be defined by implementing the interface of this
+    abstract base class.
+    """
+
+    def is_valid_dimensionality(self, d):
+        """Checks whether *d* is a valid vector dimensionality.
+
+        Parameters
+        ----------
+        d : int
+            Dimensionality
+
+        Returns
+        -------
+        bool
+            *True*, if *d* is a valid vector dimensionality for the use with
+            the algebra.
+        """
+        raise NotImplementedError()
+
+    def make_unitary(self, v):
+        """Returns a unitary vector based on the vector *v*.
+
+        A unitary vector does not change the length of a vector it is bound to.
+
+        Parameters
+        ----------
+        v : (d,) ndarray
+            Vector to base unitary vector on.
+
+        Returns
+        -------
+        ndarray
+            Unitary vector.
+        """
+        raise NotImplementedError()
+
+    def superpose(self, a, b):
+        """Returns the superposition of *a* and *b*.
+
+        This is commonly elementwise addition.
+
+        Parameters
+        ----------
+        a : (d,) ndarray
+            Left operand in superposition.
+        b : (d,) ndarray
+            Right operand in superposition.
+
+        Returns
+        -------
+        (d,) ndarray
+            Superposed vector.
+        """
+        raise NotImplementedError()
+
+    def bind(self, a, b):
+        """Returns the binding of *a* and *b*.
+
+        The resulting vector should in most cases be dissimilar to both inputs.
+
+        Parameters
+        ----------
+        a : (d,) ndarray
+            Left operand in binding.
+        b : (d,) ndarray
+            Right operand in binding.
+
+        Returns
+        -------
+        (d,) ndarray
+            Bound vector.
+        """
+        raise NotImplementedError()
+
+    def invert(self, v):
+        """Invert vector *v*.
+
+        A vector bound to its inverse will result in the identity vector.
+
+        Parameters
+        ----------
+        v : (d,) ndarray
+            Vector to invert.
+
+        Returns
+        -------
+        (d,) ndarray
+            Inverted vector.
+        """
+        raise NotImplementedError()
+
+    def get_binding_matrix(self, v, swap_inputs=False):
+        """Returns the transformation matrix for binding with a fixed vector.
+
+        Parameters
+        ----------
+        v : (d,) ndarray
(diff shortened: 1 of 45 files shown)


nengo_spa/pointer.py
```diff
@@ -21,20 +22,37 @@ class SemanticPointer(Fixed):
     rng : numpy.random.RandomState, optional
         Random number generator used for random generation of a Semantic
         Pointer.
-    vocab : Vocabulary
+    vocab : Vocabulary, optional
         Vocabulary that the Semantic Pointer is considered to be part of.
+        Mutually exclusive with the *algebra* argument.
```
Seanny123: Why different formatting for `algebra` and `vocab`?

nengo_spa/pointer.py
```diff
@@ -109,19 +122,25 @@ def __str__(self):
...
 
     def __neg__(self):
         return SemanticPointer(data=-self.v, vocab=self.vocab)
 
-    @TypeCheckedBinaryOp(Fixed)
     def __sub__(self, other):
-        type_ = infer_types(self, other)
-        vocab = None if type_ == TAnyVocab else type_.vocab
-        return SemanticPointer(data=self.v - other.evaluate().v, vocab=vocab)
+        return self + (-other)
```
Seanny123: Why does `__add__` need type checking, while `__sub__` doesn't?

jgosmann: Because sub re-expresses `a - b` as `a + (-b)` thus using the implementation of `__add__` and `__neg__` which already do type checking.

nengo_spa/algebras/vtb.py
```diff
@@ -58,10 +58,22 @@ def invert(cls, v):
         return v.reshape((sub_d, sub_d)).T.flatten()
 
     @classmethod
-    def get_binding_matrix(cls, v):
+    def get_binding_matrix(cls, v, swap_inputs=False):
```
Seanny123: I'm not sure I understand the point of swapping inputs as a function argument. Naively, it seems to me that the inputs should just be swapped in the connections and not as a parameter, if the user wants that. Would you mind explaining to me the intent of this?

jgosmann: When using a binding matrix one vector is fixed (the one the matrix) is based on and one varies over time. Thus, there is no way to swap the inputs because there is only on time varying input.

Seanny123:
I've read through the code once, but I'd like to read it again. It's pretty remarkable how you managed to abstract such a fundamental part of SPA.

nengo_spa/algebras/base.py
```diff
@@ -0,0 +1,231 @@
...
+            Inverted vector.
+        """
+        raise NotImplementedError()
+
+    def get_binding_matrix(self, v, swap_inputs=False):
+        """Returns the transformation matrix for binding with a fixed vector.
+
+        Parameters
+        ----------
+        v : n(d,) darray
```
Seanny123: What is `n` here?

Seanny123:
Made some formatting fixes and added a test to make sure the unitary vector doesn't change the length of what it's binded with. This PR LGTM.

jgosmann:
@Seanny123, can you “officially” approve this PR?

PR #201, opened 2018-07-19
Different ways to generate Semantic Pointers

### LINKED ISSUE #129: Option for orthogonal pointers in a vocabulary
The old `nengo.spa` had the option to do `spa.Vocabulary(randomize=False)` which forced all of the semantic pointers to be axis aligned and orthogonal (e.g. [1,0,0], [0,1,0] etc). I found this useful for debugging, but I don't see an option to do it in `nengo_spa`. Is there a reason it was removed, and it is straightforward to add?

jgosmann:
**Motivation and context:**
This PR adds a number of generators to generate vectors for Semantic Pointers. These can be passed to a `Vocabulary` to specify how new Semantic Pointers should be generated. This allows to easily create vocabularies with all unitary, orthogonal, or axis-aligned vectors. It also gives the user the possibility to program custom generation methods.

This removes the possibility to create a random pointer with `SemanticPointer(d)`. Not sure if this has been widely used and if some short-form syntax should be put back in. With this PR one can do `SemanticPointer(next(UnitLengthVectors(d)))`) instead.

Closes #129.

Based on #198 which introduces some changes to the code touched, but it is not essential to this PR.

**How has this been tested?**

Added and existing tests.

**Types of changes:**

**Checklist:**

**Still to do:**

- [x] Fix Python 2 compatibility.
- [ ] Shorthand for pointer generation?
- [x] Add new classes to root module?

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -31,6 +31,11 @@ Release History
   particular binding operation.
   (`#69 <https://github.com/nengo/nengo_spa/issues/69>`__,
   `#198 <https://github.com/nengo/nengo_spa/pull/198>`__)
+- Added generators for vectors with different properties that can be used to
+  define how vectors are created in a vocabulary (e.g., axis-aligned,
+  orthogonal, unitary).
+  (`#201 <https://github.com/nengo/nengo_spa/pull/201>`_,
+  `#129 <https://github.com/nengo/nengo_spa/issues/129>`_)
 - Added a matrix multiplication network ``nengo_spa.networks.MatrixMult`` based
   on the nengo-extras implementation.
   (`#198 <https://github.com/nengo/nengo_spa/pull/198>`__)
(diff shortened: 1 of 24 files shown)


jgosmann:
I'm still waiting on a review for this PR.

Seanny123:
Could there be some basic parameter validation on the `pointer_gen` argument to `spa.Vocabulary`? Would it be reasonable to check if it's an iterator or at least callable? Currently, the error is only thrown once a pointer is created.

jgosmann:
I have to look into it. We want to avoid starting the iteration (because then we'd consume an element, which is a problem especially with `AxisAlignedVectors` or `OrthogonalVectors`), but maybe it is possible to try obtaining in iterator (however there might be non-trivial differences between Python 2 and 3). Or maybe we can check for the `next`/`__next__` method on the object (not sure if all iterators have it, because there is the builtin `next`).

jgosmann:
Turns out Nengo has `is_iterable`/`is_string` functions that I could use.

Issue #184, opened 2018-06-13
Symbols accessed via vocab dictionary aren't named in Nengo GUI


Seanny123:
```
import nengo_spa as spa
import numpy as np

D = 64  # the dimensionality of the vectors
rng = np.random.RandomState(0)

num_keys = {'ONE', 'TWO', 'THREE', 'FOUR'}
color_keys = {'BLACK', 'RED', 'WHITE'}

all_vocab = num_keys | color_keys | {'NUM', 'COLOR'}

top_vocab = spa.Vocabulary(D, rng=rng, name="top")
top_vocab.populate(';'.join(all_vocab))


with spa.Network() as model:
    combined = spa.State(top_vocab)

    top_vocab['COLOR'] >> combined
```

Is it possible to have `top_vocab['COLOR']` have a labelled node in Nengo GUI?

jgosmann:
It would require to store the key (or parsed expression) in the `SemanticPointer` instance. Could be done, but not sure whether we should. The question is also what to do with operators (e.g. `vocab['A'] + vocab['B']`). The simplest solution would be to clear the name in the resulting `SemanticPointer` instance, but then the naming feature is pretty minimal. Or we could set the name to `'A+B'`, but then pretty much replicate the code from `PointerSymbol` in `SemanticPointer` and there might still be some cases where we end up without a meaningful name (at least if someone manually instantiates a `SemanticPointer` without a name and uses that in an operation).

PR #202, opened 2018-07-19
Track names of pointers from vocabulary

jgosmann:
**Motivation and context:**
Allows to show more helpful names in the GUI, addresses #184.
Note that this requires nengo/nengo_gui#991 (I could use a different symbol than `<`, but it seems like the most reasonable to me).
Complex expressions won't look super nice because parenthesis will be added around everything even if not needed (because figuring out when they are needed would make things way more complicated).

based on #198, though not strictly required

**How has this been tested?**

added tests, tried it out in the GUI

**Types of changes:**

**Checklist:**

- [ ] I have run the test suite locally and all tests passed.


jgosmann:
Also still needs a review.

Seanny123:
What code did you use to try it out in the GUI?

jgosmann:
I don't remember, probably creating some (named) `SemanticPointer` instances (or obtaining them from vocabs) and using them in expressions.

jgosmann:
Probably the last feature PR I ask you to review, @Seanny123.

Seanny123:
Using the following code:
```
import nengo
import nengo_spa as spa

dim = 32
vocab = spa.Vocabulary(dimensions=dim)

words = ['ORANGE', 'APRICOT', 'CHERRY', 'STRAWBERRY', 'APPLE']

for word in words:
    vocab.populate(word)


with spa.Network('AssociativeMemory', seed=1) as model:
    a = spa.State(vocab, label='a')
    b = spa.State(vocab, label='b')
    
    tt = spa.Transcode("ORANGE*CHERRY", output_vocab=vocab)

    
    with spa.ActionSelection() as action_sel:
        spa.ifmax(spa.dot(a, vocab["ORANGE"]+vocab["CHERRY"]),
                  a >> b)
```

The node is labeled `SemanticPointer<(ORANGE)+(CHERRY)>`.

However, if I switch the `spa.dot` to `spa,.dot(a, spa.sym.ORANGE+spa.sym.CHERRY)`, the node is no longer labeled. Is this expected?

jgosmann:
No, but the problem is slightly different than you describe and not related to this PR. The `vocab['POINTER']` notation creates a labeled node and feeds into a compare network. However, this compare network in unnecessary and the dot product could be computed in the transform for the fixed semantic pointer. When switching over to `spa.sym` notation, this is happens, i.e. the compare network and node disappear. Another (unlabeled) pass-through node appears that just transmits the dot product result (maybe it should be labeled, but this is also another problem).

If you try this on the master branch you will observe the same behaviour.

PR #209, opened 2018-08-30
Fix dot with fixed Semantic Pointer.

jgosmann:
**Motivation and context:**

Fixes the problem discovered in #202 that a dot product with a fixed `SemanticPointer` instance creates a compare network which is not needed. This ensures that a transform is used instead.

Furthermore, it fixes a problem where `spa.dot(a, b)` would return a NumPy array instead of a scalar or AST node when `a` was a `SemanticPointer`.

Somewhat related to the the confusion in #202 a label is added for the action utility node.

none

**How has this been tested?**

Extended tests except for the label of the utility node.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_spa/semantic_pointer.py
+++ b/nengo_spa/semantic_pointer.py
@@ -18,9 +18,6 @@ class SemanticPointer(Fixed):
     ----------
     data : array_like
         The vector constituting the Semantic Pointer.
-    rng : numpy.random.RandomState, optional
-        Random number generator used for random generation of a Semantic
-        Pointer.
     vocab : Vocabulary, optional
         Vocabulary that the Semantic Pointer is considered to be part of.
         Mutually exclusive with the *algebra* argument.
@@ -44,13 +41,10 @@ class SemanticPointer(Fixed):
         Name of the Semantic Pointer.
     """
 
-    def __init__(self, data, rng=None, vocab=None, algebra=None, name=None):
+    def __init__(self, data, vocab=None, algebra=None, name=None):
         super(SemanticPointer, self).__init__(
             TAnyVocab if vocab is None else TVocabulary(vocab))
         self.algebra = self._get_algebra(vocab, algebra)
-
-        if rng is None:
-            rng = np.random
 
         self.v = np.array(data, dtype=float)
         if len(self.v.shape) != 1:
--- a/nengo_spa/semantic_pointer.py
+++ b/nengo_spa/semantic_pointer.py
@@ -265,7 +265,10 @@ class SemanticPointer(Fixed):
         if isinstance(other, Fixed):
             infer_types(self, other)
             other = other.evaluate().v
-        return np.dot(self.v, other)
+        if is_array(other):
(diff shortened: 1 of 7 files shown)


jgosmann:
Is anyone able to give this a review? @Seanny123, @tcstewar?

nengo_spa/semantic_pointer.py
```diff
@@ -265,7 +265,10 @@ def dot(self, other):
         if isinstance(other, Fixed):
             infer_types(self, other)
             other = other.evaluate().v
-        return np.dot(self.v, other)
+        if is_array(other):
+            return np.dot(self.v, other)
+        else:
+            return other.dot(self)
```
arvoelke:
```
>>> a = SemanticPointer([1, 2])
>>> a.dot([1, 2])
  File "<stdin>", line 1, in <module>
  File "/home/arvoelke/CTN/nengo-spa/nengo_spa/semantic_pointer.py", line 271, in dot
    return other.dot(self)
AttributeError: 'list' object has no attribute 'dot'
```

Is this intended/acceptable?

arvoelke:
Why does the `SemanticPointer` constructor accept an `rng` argument?

nengo_spa/semantic_pointer.py
jgosmann: good catch

jgosmann:
I assume I either forgot to remove it or it got reintroduced during a merge ...

jgosmann:
Both issues should be fixed.

nengo_spa/tests/test_semantic_pointer.py
```diff
@@ -148,6 +148,9 @@ def test_dot(rng):
     a = SemanticPointer(next(gen)) * 1.1
     b = SemanticPointer(next(gen)) * (-1.5)
     assert np.allclose(a.dot(b), np.dot(a.v, b.v))
+    assert np.allclose(a.dot(b.v), np.dot(a.v, b.v))
+    assert np.allclose(a.dot(list(b.v)), np.dot(a.v, b.v))
+    assert np.allclose(a.dot(tuple(b.v)), np.dot(a.v, b.v))
```
arvoelke: Is the `other.dot(self)` code-path tested somewhere (perhaps indirectly)? Sorry if I missed this.

arvoelke: Nope. In either case, that goes to the `return np.dot(self.v, other)` code-path, since `isinstance(other, Fixed) == True` which converts `other` into an `nd.array`.

But, looking at `codecov`, the `return other.dot(self)` code-path is being hit somehow, so either I'm missing something obvious, or its being hit indirectly in some other way?

jgosmann: I see, it probably gets hit by a dot product with a `spa.Network`. `SemanticPointer` should not depend on that, so it cannot be tested in that test file. Also it is pretty straight-forward, so I think it is fine without an explicit test.

PR #214, opened 2018-09-16
Switch to new Sphinx layout.

jgosmann:
**Motivation and context:**
Switch to the new layout for the Nengo docs.

There are still some issues that need to be resolved in the nengo/nengo-sphinx-theme.

nengo/nengo-sphinx-theme#1
nengo/nengo-sphinx-theme#2
nengo/nengo-sphinx-theme#3

**How has this been tested?**
build the docs and looked through them

**Types of changes:**

**Checklist:**

**Still to do:**


tbekolay:
Just did a 0.3.0 release of nengoo-sphinx-theme, with all of the PRs that were in the queue merged, so it should be worth switching now. Let me know if there are any lingering issues with the theme blocking this PR.

jgosmann:
There are still some things I dislike in comparison to the previous theme:
* Still uses Arial which I find less legible than Lato given that many of the SPA documentation pages have quite a lot of text.
* This might be purely my preference: But the new theme uses Courier for code instead of Menlo. I find the latter to be a much nicer font.
* The heading font is too large for some of the longish page titles in the SPA documentation (e.g. [this one](https://www.nengo.ai/nengo-spa/examples/intro_coming_from_legacy_spa.html) gets broken across four lines, though maybe some of the titles should be shortened)
  * The underscore in nengo_spa on the linked example hangs too low/doesn't work well with the given line spacing.
* The code width does not fit an 80 character line (though the textwidth is fine and should not get too wide either).
* If the navigation on the left is cut-off it only scrolls once your close to the end of the page. In the old theme it would scroll immediately independent of the scroll position of the main body.
* API pages in the old theme are better structured. There you can see at a glance where a new element starts thanks to the blue background of the heading. The new theme doesn't have such highlights and the member elements of a class look the same. Furthermore, there are two awkward boxes (one around `nengo_spa.` and one around the documented element).
* The tables with “Further members” for the nengo_spa main module API documentation don't look great. the description column on the right is too narrow.
* It is not clear wether the “See also” box for the `ActionSelection` class belongs to that class or the `thalamus` member because the shape of the octagon makes it look indented (while it is not).
* There is an awkward line break in “return type” in the documentation of methods.

tbekolay:
Thanks, I made https://github.com/nengo/nengo-sphinx-theme/issues/15 to track these issues. Will let you know when they're resolved and/or ping you in for feedback if necessary.

jgosmann:
Since the theme has gotten a major update, I took another look. Most things are acceptable to me now (even though I think the theme currently used still does them a tiny bit better). Also, found some new issues:

* [ ] Awkward linebreak in some cases:
  image
* [ ] Not really a big concern to me, but given a specific content length, the two curves in the background overlap in a way where I'm not sure if that is intended:
  image

jgosmann:
@tbekolay Could you give this a quick review?

docs/conf.py
```diff
@@ -71,19 +71,16 @@
...
-html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+html_theme = 'nengo_sphinx_theme'
 html_title = "Nengo SPA {0} docs".format(release)
 html_static_path = ['_static']
 html_favicon = os.path.join('_static', 'favicon.ico')
-html_logo = os.path.join('_static', 'square-light.svg')
-html_context = {
-    'css_files': [os.path.join('_static', 'custom.css')],
-}
+html_logo = os.path.join('_static', 'full-light.svg')
```
tbekolay: The logo is handled differently in Nengo Sphinx Theme so that we don't have to include the `svg` files in every repository; set the `nengo_logo` key in `html_theme_options` to `nengo-spa-full-light.svg` (see https://github.com/nengo/nengo/blob/master/docs/conf.py#L89 as an example).

You should also set `nengo_logo_color` to #d40000 and the set `analytics_id` to be the same ID as the Nengo core one so that we can track user activity across all of the documentation page.

jgosmann: done

PR #223, opened 2019-04-26
Fix CI tests

jgosmann:
**Motivation and context:**
Set maximum pytest and xdist versions to avoid the tests failing on CI. Newer pytest versions produce warnings with Nengo 2.8 (i.e. those cannot be fixed in Nengo SPA) and newer xdist versions are incompatible with the chosen pytest version.

none

**How has this been tested?**
Waited for CI tests to pass.

**Types of changes:**

**Checklist:**

### DIFF
--- a/.ci/test.sh
+++ b/.ci/test.sh
@@ -16,7 +16,7 @@ if [[ "$COMMAND" == "install" ]]; then
     if [[ "$SCIPY" == "true" ]]; then
         conda install scipy
     fi
-    pip install 'pytest<4.0' pytest-xdist nengo=="$NENGO"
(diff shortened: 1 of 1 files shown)


jgosmann:
Actually the warnings can be dealt with (#227), however [tests than will fail](https://travis-ci.org/nengo/nengo-spa/jobs/531174227) because of the usage of `parametrize_function_name` in the `plt` fixture in Nengo. It seems to me that the fix for that will not be released in a minor/non-breaking Nengo release (I would have preferred another minor release just fixing issues like these before doing a major release, but well). Thus, limiting the pytest version seems to be the only viable option at the moment.

PR #235, opened 2019-10-19
Do not set default labels

jgosmann:
**Motivation and context:**
Given the nengo/nengo-gui#1002 has been merged, not setting any default labels will give more informative labels in the Nengo GUI. This also converts all networks to classes to be more consistent.

none

**How has this been tested?**
existing tests

Most lines did just change in indentation (due to __init__ being nested in the class) and have `net` renamed to `self`.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -28,6 +28,16 @@ Release History
 - Ensure compatibility with Nengo 3.
   (`#225 <https://github.com/nengo/nengo_spa/issues/225>`__,
   `#233 <https://github.com/nengo/nengo_spa/pull/233>`__)
+- Modules and networks no longer set a default label. This will give
+  more useful labels by default in Nengo GUI (assuming the latest dev
+  version).
+  (`#234 <https://github.com/nengo/nengo_spa/issues/234>`__,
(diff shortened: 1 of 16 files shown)


jgosmann:
It would be great to get a review on this. @tbekolay @tcstewar @Seanny123

Seanny123:
@jgosmann to make sure I understand this PR, to review, I should install the master branch of Nengo GUI and this branch of Nengo SPA. Then, I should see the better labels discussed in this PR and in the linked Nengo GUI PR?

jgosmann:
Exactly. I made some example screenshots:

Without this PR, if no explicit label is set, all State modules (for example) are just labeled "State":
image

With this PR, if no explicit label is set, the modules are labeled based on the assigned variable:
image
If the module is not assigned to a global variable, the label defaults to the class name (e.g. State).

Independent of this PR, explicitly setting the label will display that label:
image

There is also somewhat graceful degradation with older Nengo GUI versions than the current master. If no explicit label is set, but the module is not assigned to a global variable, it will be labeled "networks[i]" (where *i* is some integer).

PR #236, opened 2019-11-25
Nengo 3 compatibility

jgosmann:
**Motivation and context:**
This ensures compatibility with the recently released Nengo 3. This mainly involves removing the dependency on `nengo.utils.compat`.

Also includes some minor changes to prepare for the upcoming Nengo SPA 1.0 release.

none

**How has this been tested?**

existing unit tests

**Types of changes:**

**Checklist:**


jgosmann:
@tbekolay The build is failing because anchor links to nengo.ai are not working properly if the page isn't pre-loaded.

tbekolay:
We renamed pages so that all of the URLs that previously had underscores in them now have hyphens, so you'll have to look for references to Nengo core docs with `_` and replace them with `-`. Because we use Github pages to serve pages and therefore can't do real HTTP redirects, non-anchor links will work (because they successfully download an HTTP page that redirects to the right page) but anchor links won't work because the redirect page doesn't have those IDs.

jgosmann:
@tbekolay Fixed those problems, and switched the Nengo SPA docs over to hyphens too. Still looking good?

tbekolay:
The commits look good! The only thing that would be nice to have (but probably not essential) would be to set up redirects from the old doc page locations to the new ones. We added [a Sphinx extension](https://www.nengo.ai/nengo-sphinx-theme/extensions.html#module-nengo_sphinx_theme.ext.redirects) to Nengo Sphinx Theme to do that, but that would require switching to Nengo Sphinx Theme. It's been completely redone since last you looked at it, so it might be easier to switch now, but if that's too time consuming you can include the [redirects code](https://github.com/nengo/nengo-sphinx-theme/blob/master/nengo_sphinx_theme/ext/redirects.py) in your `conf.py`. Or not worry about it, if there aren't many things linking to those pages.

jgosmann:
I think I'm going to switch over to the Nengo Sphinx Theme which is done in PR #214, even though there are still [some things that could be improved](https://github.com/nengo/nengo-spa/pull/214#issuecomment-559218668). I will than also add the redirects (likely in a separate PR).

PR #240, opened 2020-03-22
Ensure SP algebra matches vocab

### LINKED ISSUE #239: Vocabulary doesn't consider algebra of its vectors
```
>>> import numpy as np
>>> import nengo_spa as spa
>>> from nengo_spa.algebras.vtb_algebra import VtbAlgebra
>>> vocab = spa.Vocabulary(8)
>>> v = spa.SemanticPointer(np.ones(8), algebra=VtbAlgebra())
>>> vocab.add("V", v)
>>> vocab["V"].algebra
<nengo_spa.algebras.hrr_algebra.HrrAlgebra object at 0x7f3ebb303150>
```

In this example, a vector is added with `algebra=VtbAlgebra()` but comes back as `HrrAlgebra()`. After a quick scan over the code it looks like the `algebra` attribute of the vocabulary isn't used for anything except `create_subset`. Any semantic pointers that are added/created/populated are automatically recreated using the default algebra (regardless of which algebra is passed to the vocabulary).

jgosmann:
**Motivation and context:**

Mixing algebras is not supported without explicit casts. Thus raise an exception when this is tried.

SemanticPointe.reinterpret can be used to add the pointer anyways.

Fixes #239.

**How has this been tested?**

unit test

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_spa/semantic_pointer.py
+++ b/nengo_spa/semantic_pointer.py
@@ -2,6 +2,7 @@ import nengo
 from nengo.exceptions import ValidationError
 import numpy as np
 
+from nengo_spa.algebras.base import AbstractAlgebra
 from nengo_spa.algebras.hrr_algebra import HrrAlgebra
 from nengo_spa.ast.base import Fixed, infer_types, TypeCheckedBinaryOp
 from nengo_spa.typechecks import is_array, is_array_like, is_number
@@ -63,6 +64,10 @@ class SemanticPointer(Fixed):
                 algebra = vocab.algebra
         elif vocab is not None and vocab.algebra is not algebra:
             raise ValueError("vocab and algebra argument are mutually exclusive")
+        if not isinstance(algebra, AbstractAlgebra):
+            raise ValidationError(
+                "'algebra' must be an instance of AbstractAlgebra", "algebra", algebra
+            )
         return algebra
 
     def _get_unary_name(self, op):
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,14 +68,15 @@ jobs:
         distributions: "sdist bdist_wheel "
         on:
           all_branches: true
+          tags: false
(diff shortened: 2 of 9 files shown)


nengo_spa/semantic_pointer.py
```diff
@@ -63,6 +64,10 @@ def _get_algebra(cls, vocab, algebra):
                 algebra = vocab.algebra
         elif vocab is not None and vocab.algebra is not algebra:
             raise ValueError("vocab and algebra argument are mutually exclusive")
+        if not isinstance(algebra, AbstractAlgebra):
```
jgosmann: This seems unpythonic to me. Usually duck typing should be fine and enforcing that an algebra derives from `AbstractAlgebra` technically seems like a breaking change to me.

I would prefer to leave this check out (or just test for the protocol if that can be done easily). Of course, this enables a certain class of bugs, but this is a general problem of dynamic typing and there are probably many more instances in Nengo SPA where it would apply. I'd rather consider adding type annotations across the library (though this will probably require dropping support for Python 3.4 which has reached EOL a year ago).

arvoelke: This was added because the unit test you created passed `HrrAlgebra` instead of `HrrAlgebra()` and things still worked (because the test didn't do anything with the pointers, like try to negate or add them).. so I figured if the mistake was that easy to make then something should be done to avoid having that happen again in the future.

jgosmann: I guess we should change the `AbstractAlgebra` class to a proper `ABC` which should make the `isinstance` check succeed even if the algebra does not derive from `AbstractAlgebra` as long as all required methods are implemented.

jgosmann: I made an additional commit to continue to support duck-typing for algebras, but deprecate it (so that in the future we can fully rely on the abstract base class).

PR #245, opened 2020-05-01
Add dtype to SemanticPointer and fix repr

arvoelke:
**Motivation and context:**
This is an experimental feature request (split off from #243 and rebased onto master), that allows one to experiment with other data types for semantic pointers, such as `np.complex64`.

Full support is not provided. The intended scope is for offline algebraic manipulation of SPs. In the future, one might wish to support complex types in the neural implementations as well.

Assuming #243 is merged first, then this would need to be rebased such that `dtype` also passed into the new methods for SP fractional binding. Also changes may be needed to allow `fractional_bind` to return a complex SP if the datatype is complex.

**How has this been tested?**
Unit tests pass, but new unit tests are needed (see 'Still to do').

**Types of changes:**

**Checklist:**

**Still to do:**
 - [ ] Unit test algebraic manipulation of complex SPs
 - [ ] Check all of the places where semantic pointers are created (e.g., `vocabulary.py`)
 - [ ] Add changelog entry

### DIFF
--- a/nengo_spa/semantic_pointer.py
+++ b/nengo_spa/semantic_pointer.py
@@ -28,6 +28,10 @@ class SemanticPointer(Fixed):
         with the *vocab* argument.
     name : str, optional
         A name for the Semantic Pointer.
+    dtype : type, optional
+        A datatype for the coefficients of the Semantic Pointer.
+        Defaults to `np.float64`. Complex datatypes are currently an
+        experimental feature (i.e., not fully supported).
 
     Attributes
     ----------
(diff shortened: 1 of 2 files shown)


arvoelke:
Missed another interaction with #243: for the chosen VTB fractional binding + nondegenerate vectors there was also this unit test that was stripped out:
```python
    # satisfies translational properties as long as the intermediate results
    # are allowed to be complex
    x_complex = SemanticPointer(v, algebra=algebra, dtype=np.complex128).nondegenerate()

    assert np.any(np.iscomplex((x_complex ** 0.5).v))
    assert np.allclose((x_complex ** (-0.5) * x_complex ** 2.5).v, (x * x).v)
```
but I'm not sure if there's a good reason to have this combination of features. Just stating this for reference.

PR #246, opened 2020-05-01
Truncate semantic pointer names to 1 KB

### LINKED ISSUE #244: Semantic pointer names may consume all RAM when iterating/recursing
Self-binding to a depth of 20-30 can easily consume all of the available RAM (the example here is 33MB but in my application it was consuming 38GB of RAM). Specifically, iteratively/recursively building up a semantic pointer may cause its `name` to grow exponentially in length. I accidentally 'crashed' my desktop a few times trying to figure out what was going on here. Minimal example:

```python
import nengo_spa
import numpy as np

n_iter = 20  # warning: don't set this any higher

sp = nengo_spa.SemanticPointer(np.ones(8), name="X").unitary()
for i in range(n_iter):
    sp += sp * 0.123456789
 
print("%dMB - %s..." % (len(sp.name) / (1024**2), sp.name[:128]))
```

```
33MB - (((((((((((((((((((((X).unitary())+(((X).unitary())*(0.123456789)))+((((X).unitary())+(((X).unitary())*(0.123456789)))*(0.123456...
```

arvoelke:
**Motivation and context:**
Fixes #244. 1 KB was chosen somewhat arbitrarily, but it seems long enough to capture the majority of use cases where someone might want to read the name, while being a cheap amount of memory relative to the resource requirements of the semantic pointer itself and how it will be used.

I considered adding a warning, but it didn't seem important enough to let the user know (the fact that the name is being truncated seems inconsequential for most users)?

**How has this been tested?**
Added a unit test similar to the reproducing example from #244.

**Types of changes:**

**Checklist:**

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -42,6 +42,10 @@ Release History
 - ``SemanticPointer.translate`` and ``PointerSymbol.translate`` now use the ``keys``
   argument; previously, this argument was mistakenly ignored.
   (`#248 <https://github.com/nengo/nengo_spa/pull/248>`__)
+- Fixed an issue where the names of iteratively generated semantic pointers
+  could grow exponentially in length, by truncating them at 1 KB.
+  (`#244 <https://github.com/nengo/nengo_spa/issues/244>`__,
+  `#246 <https://github.com/nengo/nengo_spa/pull/246>`__)
 
 
 1.0.1 (December 14, 2019)
--- a/LICENSE.rst
+++ b/LICENSE.rst
@@ -29,7 +29,7 @@ Licensed code
 Nengo SPA imports several open source libraries.
 
 * `NumPy <http://www.numpy.org/>`_ - Used under
-  `BSD license <http://www.numpy.org/license.html>`__
+  `BSD license <http://www.numpy.org/doc/stable/license.html>`__
 * `Sphinx <http://www.sphinx-doc.org/en/master>`_ - Used under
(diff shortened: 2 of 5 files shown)


hunse:
We may make that change in nengo-bones at some point. For now, though, I would recommend just changing the message to e.g. "This commit fixes #???" so that it's long enough.

tbekolay:
We did decide to remove the commit body minimum length, but it'll take some time before that change is done as it's a ways down our priority list.

What I would do is copy in the context that is in the [first post of this PR](https://github.com/nengo/nengo-spa/pull/246#issue-412303462) into the commit message. What's the harm in having additional context in the commit message body?

arvoelke:
> What I would do is copy in the context that is in the first post of this PR into the commit message. 

Done.

arvoelke:
> 3: B1 Line exceeds max length (285>72): "Fixes #244. 1 KB was chosen somewhat arbitrarily ...

Trying again.

arvoelke:
> `/home/travis/build/nengo/nengo-spa/docs/license.rst.rst:31:broken link: http://www.numpy.org/license.html (404 Client Error: Not Found for url: https://numpy.org/license.html)`

Trying again..

arvoelke:
> ```
> $ codespell -q 3 --skip=./build,*/_build,*-checkpoint.ipynb,./.eggs,./.git,*/_vendor --ignore-words-list=ba
> ./docs/user-guide/spa-intro.rst:129: contex ==> context
> COMMAND 'codespell' FAILED```

Trying again...

arvoelke:
It passed (unblocked)!

arvoelke:
Note: will need to drop the last two commits after #252.

PR #247, opened 2020-05-07
nondegenerate fractional vtb binding

bjkomer:
**Motivation and context:**

Expanding tests for a future version of non-degenerate VTB for fractional binding.
Currently 50% of unitary vectors are non-degenerate (corresponding to unitary vectors that form matrices with a positive determinant). Using this space instead of the householder matrix approach will permit more useful properties.
Modified `make_nondegenerate` to return a unitary vector with a positive determinant.

#243 

**How has this been tested?**

**Types of changes:**

**Checklist:**

**Still to do:**

### DIFF
--- a/nengo_spa/algebras/vtb_algebra.py
+++ b/nengo_spa/algebras/vtb_algebra.py
@@ -112,7 +112,7 @@ class VtbAlgebra(AbstractAlgebra):
 
         sub_d = self._get_sub_d(len(v))
         sub_m = fractional_matrix_power(
-            np.sqrt(sub_d) * v.reshape((sub_d, sub_d)), exponent
+            np.sqrt(sub_d) * v.reshape((sub_d, sub_d)).T, exponent
         )
         m = np.kron(np.eye(sub_d), sub_m)
         # more efficient than the equivalent:
@@ -139,7 +139,7 @@ class VtbAlgebra(AbstractAlgebra):
 
     def get_binding_matrix(self, v, swap_inputs=False):
         sub_d = self._get_sub_d(len(v))
-        m = np.sqrt(sub_d) * np.kron(np.eye(sub_d), v.reshape((sub_d, sub_d)))
+        m = np.sqrt(sub_d) * np.kron(np.eye(sub_d), v.reshape((sub_d, sub_d)).T)
         if swap_inputs:
             m = np.dot(self.get_swapping_matrix(len(v)), m)
         return m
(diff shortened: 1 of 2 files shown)


arvoelke:
Interesting find with the positive determinant restriction. I noticed that this is still using the same implementation for `fractional_bind`. Is there an advantage to rolling in the identity as it sounded you were suggesting in #243? Would that make the properties we see and the tests more self-consistent?

bjkomer:
Rolling in the identity helps somewhat, in the sense that it makes `x ** 1 == x` which is certainly more intuitive. It still doesn't make all properties consistent, since `x ** 2 != x * x`, but instead `x**2 == x *~ x`

I think ideally what would make everything more consistent is a change in the definition of what it means to bind two vectors together. By adding a transpose to the submatrices in `get_binding_matrix` binding becomes a lot more consistent with fractional powers.
Specifically changing the function to be:
```python
def get_binding_matrix(self, v, swap_inputs=False):
        sub_d = self._get_sub_d(len(v))
        m = np.sqrt(sub_d) * np.kron(np.eye(sub_d), v.reshape((sub_d, sub_d)).T)
        if swap_inputs:
            m = np.dot(self.get_swapping_matrix(len(v)), m)
        return m
```
(and also adding the transpose to the current implementation of `fractional_bind`)

This makes all of the expected properties hold
`x ** 1 == x`
`x ** 3 == (((x) * x) * x)`
`x ** 2.8 == (((x ** 1.7) * (x ** -0.4)) * (x ** 1.5))`
`x ** (a + b) == (x ** a) * (x ** b)`
etc

I understand this will be a breaking change, not in terms of API but in terms of neural models created using VTB. I'm not sure what all of the implications will be, @jgosmann probably has a better idea.

The only other alternative I see would be to come up with a different way of implementing fractional binding, but right now I don't see a way of getting a solution where
`x ** 1 == x` and `x ** 2 == x * x` 
for the current binding implementation.

bjkomer:
Another consequence of this binding re-definition is that the identity vector becomes a true identity, rather than just a right-identity. `A * I == I * A == A`

nengo_spa/algebras/vtb_algebra.py
```diff
@@ -112,32 +112,34 @@ def fractional_bind(self, v, exponent):
...
         )
         m = np.kron(np.eye(sub_d), sub_m)
         # more efficient than the equivalent:
         # m = fractional_matrix_power(self.get_binding_matrix(v), exponent)
         return m.dot(self.identity_element(len(v)))
 
     def make_nondegenerate(self, v):
-        # note: these "Householder vectors" are not only unitary, but symmetric
+        # nondegenerate vectors are unitary and
+        # form a matrix with a positive determinant
```
jgosmann, line 124: This seems to be wrong? `np.linalg.det(-np.eye(d))` is 1, but `fractional_matrix_power(-np.eye(d), 0.5)` gives complex values.
I think the relevant property for (T)VTB is positive definiteness.

jgosmann:
Superseded by #271.

PR #259, opened 2020-06-22
Fix SPA sequence example

### LINKED ISSUE #258: SPA sequence example is broken
The [SPA sequence](https://www.nengo.ai/nengo-spa/examples/spa-sequence.html#) is not really producing a sequence as expected. Increasing dimensionality or changing the seed did not fix it.

arvoelke:
The stimulus wasn't connected to the state being sequenced, and so it would randomly break depending on whether the initial state was close enough to some vector in the vocabulary.

**Motivation and context:**
Fixes #258.

**How has this been tested?**
spa-sequence

Ran it locally and tried out a bunch of different seeds and longer simulation times. It always seems to do the right thing now.

**Types of changes:**

**Checklist:**
- [N/A] I have updated the documentation accordingly.
- [x] (added by Jan) I have included a changelog entry.
- [N/A] I have added tests to cover my changes.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -57,6 +57,10 @@ Release History
   `SemanticPointer` instances.
   (`#250 <https://github.com/nengo/nengo_spa/issues/250>`__,
   `#257 <https://github.com/nengo/nengo_spa/pull/257>`__)
+- The SPA sequence example was not working properly due to a missing connection.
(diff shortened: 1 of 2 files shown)


arvoelke:
I don't see the changelog entry here (just the same commit as before, with a different hash, that has been force-pushed) in case that was to show up here.

jgosmann:
Yeah, I forgot to add the changes to the index on my first try, but is already fixed.

PR #264, opened 2020-08-23
Add dependency version constraints

jgosmann:
**Motivation and context:**
Add version constraint to all dependencies to prevent things from breaking randomly (as long as dependencies follow Semantic Versioning), especially in CI builds. Also, allow pytest 6 to fix incompatibility with pytest-xdist 2 in the CI builds.

~This supersedes commit 51bf91b5d686851954d37da8ef64038283f86d57 from PR #263.~

**How has this been tested?**

Existing tests and building documentation with minimum versions of dependencies and maximum available versions ~satisfying the constraints~.

**Types of changes:**

**Checklist:**

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,20 +30,21 @@ version_module = imp.load_source(
 )
 testing = "test" in sys.argv or "pytest" in sys.argv
 
+numpy = "numpy>=1.8"
 docs_require = [
-    "jupyter_client",
-    "sphinx",
-    "nbsphinx",
-    "nengo_sphinx_theme",
+    "jupyter_client>=5.3.4",
+    "sphinx>=2.2.1",
+    "nbsphinx>=0.5.0",
+    "nengo_sphinx_theme>=1.2.0",
 ]
-optional_requires = ["scipy"]
+optional_requires = ["scipy>=1.4.1"]
 tests_require = [
-    "jupyter",
+    "jupyter>=1.0.0",
     "matplotlib>=2.0",
-    "nbformat",
+    "nbformat>=5.0.7",
     "pytest>=3.6",
-    "pytest-plt",
-    "pytest-rng",
+    "pytest-plt>=1.0.1",
+    "pytest-rng>=1.0.0",
 ]
 
 setup(
@@ -59,8 +60,8 @@ setup(
     long_description=read("README.rst", "CHANGES.rst"),
     zip_safe=True,
     include_package_data=True,
-    setup_requires=["pytest-runner"] if testing else [] + ["numpy>=1.8",],
-    install_requires=["nengo>=2.7,<4", "numpy>=1.8",],
+    setup_requires=["pytest-runner>=5.2"] if testing else [] + [numpy,],
+    install_requires=["nengo>=2.7", numpy,],
     extras_require={
         "all": docs_require + optional_requires + tests_require,
         "docs": docs_require,
--- a/setup.cfg
+++ b/setup.cfg
@@ -44,6 +44,7 @@ markers =
 filterwarnings =
     ignore::ImportWarning
     ignore:(Buffer|Memory):DeprecationWarning
+    ignore:IPython.core.inputsplitter is deprecated:DeprecationWarning
     ignore:Skipping some optimization steps
     ignore:SciPy is not installed
     ignore:numpy.(dtype|ufunc) size changed


jgosmann:
> but upper bounds are going to cause conflicts with other projects eventually.

I don't see in how far this will be causing conflicts. Either the intersection of valid dependency versions with other projects is non-empty and the projects + dependencies can be installed with tested versions or the intersection is empty. In the latter case nothing will be installed first, but there is still the option of to force install with full knowledge that things might no work. Without upper bounds you might get a non-working installation without an obvious reason.

> You're also making the assumption that other projects are using semantic versioning (and using it correctly); they could just as easily break something in a minor release.

Yes, but most projects either state explicitly to follow semantic versioning or at least seem to do. Even if not, a change in major version seems like something I want to explicitly test before declaring it as compatible. For all other version changes (except nbsphinx which is <1.0), everything is as before. Sure, a minor version increment can also break things because mistakes happen, but in that case things break either way. I think there is still benefit from preventing the automatic usage of new major versions that are much more likely to break things.

> The reason we don't bother with exact version requirements in our CI is because we want to find those incompatibilities before users do.

Finding alone doesn't help without updating the code or introducing an upper constraint later. I'm doing this in my free time and don't always have time to fix things until days or maybe weeks later. I rather prefer version constraints that ensure that users get a working installation automatically and then I can bump versions when I've got the time. (And upgrades for versions that don't break anything could likely be automated.)

There's also a somewhat anecdotal reason why I think upper version bounds are a benefit. I recently had to deploy a Python application with various dependencies and transitive dependencies on a system that was only supporting an "old" Python version (2.7? 3.4? Can't remember.) Of course, some of the dependencies weren't compatible with that Python version anymore, so an older version had to be used. But that older version wouldn't work with the newest transitive dependency, so one has to figure out the latest version of the transitive dependency that still works with the older version. That is significantly easier if there are already upper version bounds defined. Otherwise, you're sort of stuck with testing every single version.

Thinking about that and coming back to the beginning, it seems to me that not having upper bounds will cause conflicts. Essentially, you loose the possibility to easily install older versions because dependencies will be installed in their newest version and will probably break at some point. Being able to install an old Nengo SPA version without a lot of effort to figure out matching dependencies seems actually pretty essential to me. I'm not continuously updating my Nengo SPA models for new to use new versions (and I don't expect other people to do that with their models), but it should remain possible to easily run models created with a specific version of Nengo SPA in the future by just installing that particular version.

setup.py
```diff
@@ -30,20 +30,21 @@ def read(*filenames, **kwargs):
...
+optional_requires = ["scipy>=1.4.1,<2"]
 tests_require = [
-    "jupyter",
-    "matplotlib>=2.0",
-    "nbformat",
-    "pytest>=4.0,<5",
-    "pytest-plt",
-    "pytest-rng",
+    "jupyter>=1.0.0,<2",
+    "matplotlib>=2.0,<3",
```
jgosmann, line 43: - [ ] Just noticed that there is already a version 3 of matplotlib, I should check whether that works fine.

tbekolay:
> I don't see in how far this will be causing conflicts.

There aren't conflicts if 100% of all projects are 100% vigilant in making sure everything stays backwards compatible within the major version (which, again, not all projects claim to do). But there are lots of situations where you can get burned unnecessarily.

- Say a new feature is introduced in Matplotlib 2.12 that requires a bump in the NumPy version; if you're requiring a version less than that version, then you're in a situation that is impossible to satisfy because Matplotlib requires something that you are saying in incompatible with NengoSPA.
- Say a package introduces a new dependency, or you simply missed one of the dependencies that one of your upstream packages require. Since you're not limiting the upper version of those, you might be allowing them to upgrade, but those upgraded versions are incompatible with the upstream version that you're forcing. E.g., say the `cycler` package that Matplotlib depends on upgrades, and does so in an incompatible way with older Matplotlib versions. Since you allow the `cycler` upgrade, 
  - You might think that this means you should do all the dependency resolution manually and add all possible dependent packages to your setup.py with upper bounds. You could maybe do this, and it would work for a bit until some package introduces a new dependency. But your goal was to not have to keep up with all this development, so I don't really see the point. Also, it would inflate the size and complexity of your setup.py dramatically. Imagine if all projects on pypi did this, it would be untenable.

In addition to these actual conflicts, you're also limiting the freedom a user has in their isolated environments. Say NumPy makes some nice optimizations in a future version. If they are using NengoSPA with an upper bound then they can no longer use those optimizations because you say that it is incompatible with NengoSPA. It is possibly true (for now) that a user can force the upgraded installation, but that might not be the case in the future. Also, suppose they have some issue and uninstall and reinstall NengoSPA. Then you've forcibly downgraded their NumPy install, losing those optimizations, and possibly in a way that the user doesn't notice, their models just get slower because they installed NengoSPA.

> Without upper bounds you might get a non-working installation without an obvious reason.

It's possible that reasons could be non-obvious, but in my experience it's relatively straightforward to debug and fix an issue related to a new version of a package.

> most projects either state explicitly to follow semantic versioning or at least seem to do

It seems like a big investment of time to figure out whether each project uses semantic versioning, and does it correctly.

> I think there is still benefit from preventing the automatic usage of new major versions that are much more likely to break things.

I'm curious if you've seen any other Python projects do this? If it were indeed a useful thing, I would expect it to be common. But I've never seen it except in specific circumstances with specific dependencies.

>  Essentially, you loose the possibility to easily install older versions because dependencies will be installed in their newest version and will probably break at some point. ... it should remain possible to easily run models created with a specific version of Nengo SPA in the future by just installing that particular version.

This just isn't the kind of use case that is realistic to optimize for. Everything changes over time, and it will never be the case that you can fully get back some old environment unless you guard against that when you made the environment in the first case. `pip` itself has been making big updates lately too; it might be the case that no matter how many things you do to try to keep things historically recoverable, a `pip` update will make it all for naught, or maybe some older package requires a version of Python that is no longer supported and therefore not packaged for the operating system you're using. If keeping something in a certain state is extremely important for a some reason, then a better solution is likely going to involve an operating system image and a docker container, or something like that.

jgosmann:
>     * Say a new feature is introduced in Matplotlib 2.12 that requires a bump in the NumPy version; if you're requiring a version less than that version, then you're in a situation that is impossible to satisfy because Matplotlib requires something that you are saying in incompatible with NengoSPA.

Wouldn't this be satisfiable by staying on Matplotlib 2.11 until NengoSPA has bumped the upper version bound?

>     * Say a package introduces a new dependency, or you simply missed one of the dependencies that one of your upstream packages require.

I don't think that transitive dependencies should be added (with or without version constraints), except maybe in very exceptional cases. If a specific version of a library does not install because one of their dependencies updated, it is a bug in that library to me. (And just because a majority of the Python ecosystem is broken in this way, it is not an argument to proliferate that.)

> If they are using NengoSPA with an upper bound then they can no longer use those optimizations because you say that it is incompatible with NengoSPA.

Only until the next release which should follow in a somewhat timely manner to bump the bound. (Which should be automatable to a large degree ...) I'm not intending to keep this upper bounds fixed indefinitely.

> Then you've forcibly downgraded their NumPy install

That's actually true and a large flaw of `pip`, hopefully resolved by the new resolver ... (no pun intended)

> It seems like a big investment of time to figure out whether each project uses semantic versioning, and does it correctly.

This doesn't have to be 100% perfect. No matter what versioning scheme is used, it seems to me a pretty reasonable assumption that any bump in major version implies large changes that are likely to break things and should be explicitly checked for compatibility. For projects that make a strong commitment to backwards compatibility the upper bound could be skipped. And sure, things can break with minor version updates too, but it's not a black and white thing. It's a trade-off between false positive and negatives in the compatibility.

> I'm curious if you've seen any other Python projects do this? If it were indeed a useful thing, I would expect it to be common. But I've never seen it except in specific circumstances with specific dependencies.

I think it is less common in the Python community for historic reasons. The tooling hasn't been great (or still isn't), but it is improving (e.g. [Poetry](https://python-poetry.org/), which is also arguing that [versions should not be unbounded](https://python-poetry.org/docs/faq/#why-are-unbound-version-constraints-a-bad-idea)). I think being able to automate the upper version bounds is also an important puzzle piece which hasn't been there for forever.

MyPy seems to have upper version bounds, [Docker Compose](https://github.com/docker/compose) and [Strawberry](https://github.com/strawberry-graphql/strawberry) too. [Others also argue pro upper version bounds.](https://blog.miguelgrinberg.com/post/the-package-dependency-blues)

In the JavaScript/NPM world and for Rust/cargo upper version bounds and the assumption of Semantic Versioning are pretty much standard. It seems to work there.

> Everything changes over time, and it will never be the case that you can fully get back some old environment unless you guard against that when you made the environment in the first case.

Again, this is not black and white. Even if things still break, having lower and upper version boundaries at least reduces the search space significantly.

Anyways, to me it's not worth to continue fighting for, so I guess I will remove the upper bounds.

tbekolay:
> Wouldn't this be satisfiable by staying on Matplotlib 2.11 until NengoSPA has bumped the upper version bound?

I was assuming a new installation from scratch, which is what happens during CI testing.

> That's actually true and a large flaw of `pip`, hopefully resolved by the new resolver

:crossed_fingers: 

> MyPy seems to have upper version bounds, [Docker Compose](https://github.com/docker/compose) and [Strawberry](https://github.com/strawberry-graphql/strawberry) too. [Others also argue pro upper version bounds.](https://blog.miguelgrinberg.com/post/the-package-dependency-blues)

Thanks for the links to the other projects with upper bounds and the links to blog posts; I hadn't ever seen anyone using or arguing for upper bounds before, so it's definitely food for thought. I think if we were to restrict major version upgrades, we would want to do it across the ecosystem and probably with some infrastructure help from NengoBones. But maybe it is something that we will want to do, especially coupled with adopting a more robust build system like Poetry. We'll add it to our backlog.

jgosmann:
@tbekolay Not sure if you got a notification, but I've updated the PR.

jgosmann:
ping @tbekolay

jgosmann:
@tbekolay Do you still want to give this a review after my update? It's been lying around for quite a while and would like to merge this soon to release to make a 1.1.1 release.

tbekolay (approved):
LGTM :+1:

PR #265, opened 2020-08-31
Support to differentiate the sidedness of algebra special elements

jgosmann:
**Motivation and context:**
So far NengoSPA assumed that all special elements (i.e. identity, zero, absorbing element, and inverse) of an algebra are two-sided, i.e. can be applied to either side in the binding with the same effect. While this is true for HRR/circular convolution, it does not hold in general. In particular, VTB only has a right identity and inverse(*), and by transposing the VTB-matrix a variant is obtained where there are only a left identity and inverse.

This PR changes things, so that one can explicitly ask for the left, right or two-sided inverse. In most methods this is done via a new `sidedness` argument taking a corresponding enum value. For backwards compatibility this argument is `ElementSidedness.TWO_SIDED` by default. Furthermore, for VTB this value will return the right identity or right inverse, but produce a deprecation warning. With Semantic Pointers `~` is used as inversion operator and it is not possible to add an additional argument here. Thus, `~` will be explicitly for the two-sided inverse. It still works with VTB, but again will produce a deprecation warning. For VTB SPs `.rinv()` can be used and there is `.linv()` for symmetry.

(*) Interestingly, it is still possible to unbind from the left, but this is not done by obtaining an inverse of the the left operand, but by applying the swapping matrix to the right (bound) vector and then using the right inverse from the right side to unbind.

The necessity for this changes became evident during discussing PR #243, in particular [this thread](https://github.com/nengo/nengo-spa/pull/243#discussion_r429616559). It has not been obvious that VTB identity vectors and inverses are only right-sided.

**How has this been tested?**

Additional and existing tests. Some changes where necessary to handle missing special elements of one side in the generic tests, as well as some handling of the deprecation warnings where behaviour deviates from what sidedness is requested for backwards compatibility.

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_spa/tests/test_semantic_pointer.py
+++ b/nengo_spa/tests/test_semantic_pointer.py
@@ -92,7 +92,7 @@ def test_add_sub(algebra, rng):
 
 
 @pytest.mark.parametrize("d", [64, 65])
-def test_binding_and_inversion(algebra, d, rng):
+def test_binding(algebra, d, rng):
     if not algebra.is_valid_dimensionality(d):
         return
 
@@ -100,7 +100,6 @@ def test_binding_and_inversion(algebra, d, rng):
 
     a = SemanticPointer(next(gen), algebra=algebra)
     b = SemanticPointer(next(gen), algebra=algebra)
-    identity = Identity(d, algebra=algebra, sidedness=ElementSidedness.RIGHT)
 
     c = a.copy()
     c *= b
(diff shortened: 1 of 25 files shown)


jgosmann:
@tbekolay https://uwspace.uwaterloo.ca/handle/10012/13498 is reported as broken link due to some certificate issue, but my local browser says the certificate is fine.

```
sphinx.errors.SphinxWarning: /home/travis/build/nengo/nengo-spa/docs/modules/nengo_spa.algebras.rst.rst:43:broken link: https://uwspace.uwaterloo.ca/handle/10012/13498 (HTTPSConnectionPool(host='uwspace.uwaterloo.ca', port=443): Max retries exceeded with url: /handle/10012/13498 (Caused by SSLError(SSLCertVerificationError(1, '[SSL: CERTIFICATE_VERIFY_FAILED] certificate verify failed: unable to get local issuer certificate (_ssl.c:1051)'))))
```

nengo_spa/tests/test_semantic_pointer.py
```diff
@@ -307,22 +341,41 @@ def test_invalid_algebra():
     SemanticPointer(next(gen), algebra=HrrAlgebra())
 
 
-def test_identity(algebra):
-    assert np.allclose(Identity(64, algebra=algebra).v, algebra.identity_element(64))
+@pytest.mark.filterwarnings("ignore:.*:DeprecationWarning")
+@pytest.mark.parametrize("sidedness", ElementSidedness)
```
Seanny123, line 345: Could you help me understand why `ElementSideness` is a `parametrize` parameter? It seems like a constant that could be just defined in the function?

jgosmann: Depending on the `algebra` (which is also parametrized), there might only be identities for specific sides. Left and right identity might also not be the same in the general case (though currently they are for `HrrAlgebra` and `VtbAlgebra` does only have a right identity).

Seanny123: Oh, I didn't understand that `ElementSidedness` meant it would iterate through each value of the `Enum` as a parameter.

jgosmann:
> @tbekolay https://uwspace.uwaterloo.ca/handle/10012/13498 is reported as broken link due to some certificate issue, but my local browser says the certificate is fine.

Apparently the server is misconfigured and does not send the certificate chain. I couldn't think of anything better than changing it to http instead of https.

jgosmann:
Switching to http seemed to have worked locally, but on Travis-CI it's complaining again. 😠

PR #266, opened 2020-09-05
Implement TVTB algebra

jgosmann:
**Motivation and context:**

Transposed Vector-derived Transformation Binding (TVTB) uses the same binding method as VTB, but transposes the derived matrix. This makes the identity and inverse two-sided.

Based on #265 and PR is against that branch to get the relevant diff.

**How has this been tested?**

Mostly existing generic algebra tests, but tests for the VTB network have been copied and used to test the TVTB network.

I'd suggest to take a look at `TvtbAlgebra` first (compare it to `VtbAlgebra`) and then look at the `TVTB` network (and compare it to the `VTB` network).

**Types of changes:**

**Checklist:**

### DIFF
--- a/nengo_spa/algebras/tests/test_algebras.py
+++ b/nengo_spa/algebras/tests/test_algebras.py
@@ -52,7 +52,7 @@ def test_binding_and_invert(algebra, d, sidedness, rng):
                 warnings.simplefilter("error", DeprecationWarning)
                 if binding_side is ElementSidedness.LEFT:
                     bound = algebra.bind(b, a)
-                    r = algebra.bind(bound, algebra.invert(b, sidedness=sidedness))
+                    r = algebra.bind(algebra.invert(b, sidedness=sidedness), bound)
                 elif binding_side is ElementSidedness.RIGHT:
                     bound = algebra.bind(a, b)
                     r = algebra.bind(bound, algebra.invert(b, sidedness=sidedness))
@@ -82,8 +82,10 @@ def test_get_binding_matrix(algebra, rng):
     b = next(gen)
 
     m = algebra.get_binding_matrix(b)
-
     assert np.allclose(algebra.bind(a, b), np.dot(m, a))
+
+    m = algebra.get_binding_matrix(b, swap_inputs=True)
+    assert np.allclose(algebra.bind(b, a), np.dot(m, a))
 
 
 @pytest.mark.filterwarnings("ignore:.*sidedness:DeprecationWarning")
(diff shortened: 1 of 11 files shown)


nengo_spa/algebras/tests/test_algebras.py
```diff
@@ -52,7 +52,7 @@ def test_binding_and_invert(algebra, d, sidedness, rng):
                 warnings.simplefilter("error", DeprecationWarning)
                 if binding_side is ElementSidedness.LEFT:
                     bound = algebra.bind(b, a)
-                    r = algebra.bind(bound, algebra.invert(b, sidedness=sidedness))
+                    r = algebra.bind(algebra.invert(b, sidedness=sidedness), bound)
```
Seanny123, line 55: I'm guessing this was an error that wasn't evident until TVTB and it's sidedness was implemented?

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -0,0 +1,206 @@
...
+        sub_d = np.sqrt(d)
+        return sub_d * sub_d == d
+
+    def _get_sub_d(self, d):
+        sub_d = int(np.sqrt(d))
+        if sub_d * sub_d != d:
+            raise ValueError("Vector dimensionality must be a square number.")
+        return sub_d
+
+    def make_unitary(self, v):
```
Seanny123, line 84: Is there a name or explanation for this algorithm?

nengo_spa/algebras/tests/test_algebras.py
jgosmann: Yes, the left-inverse needs to be applied from the left.

nengo_spa/algebras/tvtb_algebra.py
jgosmann: Maybe ... probably? It's generating an orthogonal matrix, same algorithm as the original VTB.

jgosmann: Is it the [Gram-Schmidt process](https://en.wikipedia.org/wiki/Gram%E2%80%93Schmidt_process)?

Seanny123: I think I even left the same comment when reviewing the original VTB...

Once again, upon further consideration, I think I can trust the name `make_unitary` even if I don't completely grasp the internals.

Seanny123 (approved):

PR #270, opened 2020-11-08
Generate better SemanticPointer names

### LINKED ISSUE #255: Generate better names
Currently the generated names for Semantic Pointers contain more parentheses than strictly necessary. Only using parentheses were necessary would improve readability. This will require to give the name generation code an understanding of the structure so that it can determine operator precedence and derive whether parentheses are necessary.

jgosmann:
**Motivation and context:**

Instead of immediately constructing a string for the SemanticPointer
name, this uses the expr_tree classes to construct a syntax tree.
This enables keeping track of operation precedence and allows to
render a name with only the necessary parenthesis increasing the
readability of SemanticPointer names.

Furthermore, this implements a visitor for this syntax tree to limit the
rendered length while preserving correctness of the parenthesis.

Closes #255.

none

**How has this been tested?**

* added unit tests
* modified the test for maximum name length accordingly

**Types of changes:**

**Checklist:**

**Still to do:**

* [x] Handle `KeywordArgument` when limiting length

### DIFF
--- a/nengo_spa/semantic_pointer.py
+++ b/nengo_spa/semantic_pointer.py
@@ -106,7 +106,7 @@ class SemanticPointer(Fixed):
         if isinstance(other, SemanticPointer):
             other_expr_tree = other._expr_tree
         else:
-            other_expr_tree = Leaf(str(other))  # FIXME what is this for?
+            other_expr_tree = Leaf(str(other))
         self_expr_tree = self._expr_tree
         if self_expr_tree and other_expr_tree:
             if swap:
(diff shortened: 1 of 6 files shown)


nengo_spa/semantic_pointer.py
```diff
@@ -78,21 +93,25 @@ def _get_algebra(cls, vocab, algebra):
...
 
     def _get_binary_name(self, other, op, swap=False):
         if isinstance(other, SemanticPointer):
-            other_name = other.name
+            other_expr_tree = other._expr_tree
         else:
-            other_name = str(other)
-        self_name = self.name
-        if self_name and other_name:
+            other_expr_tree = Leaf(str(other))  # FIXME what is this for?
```
Seanny123, line 109: Do you want to resolve this before merging?

jgosmann: oops 😶 Removed the comment, wrote it in a confused moment, but the `other` operand migth not be a `SemanticPointer` (could be a scalar for example), thus this is required.

Seanny123 (approved):
I read this MR as carefully as I could and I think it's solid. Even if I couldn't grasp how each unit-test case was necessary/sufficient.

jgosmann:
Thanks a lot!

PR #271, opened 2020-11-15
Implement signs and binding powers for algebras

jgosmann:
# Motivation and context

## Signs

This implements the notion of "signs" for algebras. The information about a sign allows to tell wether a vector will change the subspace of another vector under binding. In particular, fractional binding powers are only possible with a "positive" vector (at least if a real-valued result is desired).

Also see [this comment with subsequent discussion](https://github.com/nengo/nengo-spa/pull/243#issuecomment-721942995).

I'm mostly using the terms "positive" and "negative", but potentially it would be more appropriate to speak of "positive definite" and "negative definite". For the (T)VTB algebras the concepts directly relate to the definiteness of the binding matrix, but for HRR they do not. Not sure if either pair of terms in more "general".

## Binding powers

Furthermore, this implements binding powers, in particular, fractional binding powers. There are some differences with regards to #243:

1. For negative exponents, the inverse is explicitly taken (instead of relying on the exponentiation of the Fourier coefficients for HRR or the fractional matrix power for (T)VTB). For unitary vectors either method should produce the same result. For non-unitary vectors `v`, `v ** -1` would than give the approximate inverse (which is consistent with the notion used in some papers if I recall correctly). However, the relation `(v**a)**b == v**(a*b)` won't hold anymore for non-unitary vectors if negative exponents are involved. I'm undecided which variant might be better. 🤔
   (In the original PR it was commented that this relation does not hold in general anyways because sqrt is multi-valued. But with this rationale fractional binding powers couldn't be a thing because `v**0.5` itself would already be multi-valued. The reason this is possible because the "positive" result is picked by convention, even though the "negative" one could be taken as well.)
2. Instead of checking whether the final result contains imaginary components (which relies on some error margin around 0), I check the "sign" prior to the calculation. I think this should be equivalent (given the signs are correctly defined).
3. The VTB implementation of (fractional) binding powers is somewhat hacky, but I think this is unavoidable because we have no left identity that can be used as "base" value. Thus, the exponentiated vector itself needs to be used as "base" value requiring to adjust the exponent accordingly, account for it being potentially zero etc. Swapping the role of the operands in the binding operator, does not solve this. While it provides a left (instead of a right) identity, the binding power would no longer result in a matrix power, leaving us with "manually" applying the binding for each value in `range(exponent)`. This eliminates the possibility for fractional binding powers.
4. Due to these issues the relations `(v**a)**b == v**(a*b)` and `(v**a)*(v**b) == v**(a+b)` do not hold for the VTB. If you really want to use fractional binding powers, you probably do not want to use VTB, but HRRs or TVTB.
5. I think both methods to generate "non-degenerate" vectors for (T)VTB in #243 and #247) are not actually achieving what is desired. I think the proper condition is to have a vector that gives a positive definite binding matrix. The method in #243 is somewhat close (at least it generate a symmetric matrix which is a requirement for positive definiteness), but the eigenvalues of the Householder matrix are from {-1, 1}, i.e. can be negative which violates positive definiteness. I tested that Householder matrices with negative eigenvalues produce complex values under fractional binding. The method in #247 only forces the determinant to be positive, but this does not imply positive eigenvalues/positive definiteness. Instead I implemented the following methods to generate a non-degenerate/positive vector:
  * For non-unitary vectors with binding matrix m: `sqrtm(m @ m.T)` (basically squaring and taking the square root to obtain an absolute value)
  * For unitary vectors, the only unit-length, real-valued, positive vector is the identity.

## Complex numbers

By allowing complex values in the Semantic Pointers, fractional binding could be extended to be valid in more cases (the same way complex numbers allow exponents of negative numbers). But at the moment I don't see the added complexity worth it. Also because it would require more complicated neural implementations.

## Vector creation API

I decided against a `make_nondegenerate` method:

* `make_unitary` and `make_nondegenerate` are not commutative, so a third method `make_nondegenerate_unitary` would be required. If we get more methods, this could quickly lead to a combinatorial explosion.
* I don't see a very strong use case to generate a non-degenerate vector that is close to an existing vector, especially as it won't be that close most of the time. I suppose it most use cases generating a random vector with the desired property is sufficient.

The new `create_vector` method allows to specify a set of properties that the returned vector needs to fulfill avoiding the combinatorial explosion (at least on the API side, implementation-wise this might still happen depending on the algebra, but isn't really avoidable).

I do not intend to deprecate the `make_unitary` method at the moment though.

## Higher level APIs

Currently, this PR is only concerned with the low-level API of algebras. I think this is sufficient for this PR and tackling more in it would make it too large and distract from the issues of getting the math right. There will be follow up PRs making this functionality accessible via the higher-level APIs of `SemanticPointer` and `vector_generation`.

# Interactions with other PRs:

Supersedes #243 and #247.

# How has this been tested?

Added tests.

* [x] Probably should run each of these tests a couple of times to ensure that they do not only pass because we got lucky with the random vector used.

**Types of changes:**

**Checklist:**

- [x] I have updated the documentation accordingly. (API documenation only, for user guide see PR #282.)

**Still to do:**

* [x] Methods to create "positive" vectors. Currently I am undecided how the API should look like, suggestions welcome. It needs to be taken into account:
  * We might have multiple valid methods to create a "positive" vector. For HRR this would be the method proposed in #243 as well as the already implemented `abs` method. (Note that the `abs` method does not work on indefinite vectors. The (T)VTB vectors are in general indefinite because they are usually not symmetric. Thus, the `abs` method cannot be used in general there.)
    * Should multiple methods be offered in that case? Or should a single canonical way be defined?
  * We might need to use different methods depending on the vector. The method in #243 for (T)VTB produces a unitary positive vector, but it might be desirable to create positive vectors that are not unitary. The latter can be generated by multiplying the transpose of the binding matrix to itself and then calculating the fractional binding power of 0.5 and converting the result back to a vector. This method is, however, not suitable for unitary vectors that would reduce to the identity with this method.
  * A `make_unitary` and `make_positive` will in general not commute, i.e. the property obtained from the first operation might be lost through the second operation.
  * The `make_unitary` API exists in this way probably for two reasons:
    * We need to create a normal random vector anyways before we can make in unitary. We cannot generate a unitary vector directly? (I suppose we probably could in the Fourier space, couldn't we?)
    * We want a unitary vector that is close to the original vector. (Though how much similarity is actually preserved?)
    Potentially, it makes more sense to have `create_unitary`, `create_...` functions that do not depend on an existing vector. In particular, for (T)VTB the original vector does seem to be very relevant to the final vector and is more or less just used as a source of random numbers.
* [x] Implement and test fractional binding powers.
* [x] More testing of signs utilizing methods to create positive vectors and fractional binding powers?
* [x] Consider numerical stability with HRR and FFT coefficients close to π (direction of rotation depends on rounding).
* [x] Add reference to (done in PR #282, not here)
  ```
    .. [komer2019] Komer, B., Stewart, T.C., Voelker, A.R. and Eliasmith, C.
       A neural representation of continuous space using fractional binding.
       Proceedings of the 41st Annual Meeting of the Cognitive Science
       Society. 2019.
  ```


jgosmann:
@bjkomer @arvoelke (maybe also @tcstewar?)
Please take a look, double check the math, and comment what you think.

jgosmann:
@bjkomer @arvoelke @tcstewar 
What is your interest level in this? I know @arvoelke commented elsewhere that he won't have time to look further into this; which is fine of course. But if no one has time or interest in bringing this forward, I have to consider whether I still want to continue work on this and if I am comfortable to merge what is probably the largest feature since version 1.0 without a proper review.

nengo_spa/algebras/hrr_algebra.py
```diff
@@ -82,6 +126,48 @@ def bind(self, a, b):
...
+        --------
+        .sign
+        """
+        if int(exponent) != exponent and not self.sign(v).is_positive():
+            raise ValueError(
+                "Fractional binding powers are only supported for 'positive' vectors."
+            )
+        if exponent < 0:
+            v = self.invert(v)
+        return np.fft.irfft(np.fft.rfft(v) ** abs(exponent))
```
arvoelke, line 169: Quick note: with `irrft` I think you need to pass `n=len(v)` to handle the case where `n` is odd? Just going off memory.

arvoelke:
Hey @jgosmann. :smiley: @pblouw or @bjkomer might have some time to look into this soon, and if not I should be able to get to it first thing in March. We're all interested in this and it's on our radar. Thanks for the ping.

jgosmann:
Ok, sounds good. No rush, just wanted to know the general level of interest.

nengo_spa/algebras/hrr_algebra.py
arvoelke: I added some odd dimensions to a couple unit tests and it failed, so I made this fix and then they passed. Pushed that commit so marking this resolved. Still need to check that the notion of signs are being handled appropriately for both odd and even case (https://github.com/nengo/nengo-spa/pull/243#issuecomment-723260166).

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -1,7 +1,14 @@
+import warnings
```
arvoelke, line 1: Minor thing but the commit that touches all the `tvtb_algebra.py` stuff is titled "Implement fractional binding powers for VtbAlgebra" when it should be "Implement fractional binding powers for TvtbAlgebra" but that pushes the character limit in the commit name from 50 characters over the limit to 51 characters.

arvoelke:
All the commits up until now and the rebasing onto #279 have been to get the PR into a state where it is passing the CI. Now I'll actually start reviewing this.

docs/modules/nengo_spa.algebras.rst
```diff
@@ -13,6 +13,7 @@ nengo\_spa\.algebras
       CommonProperties
       ElementSidedness
       HrrAlgebra
+      HrrProperties
```
arvoelke, line 16: I assume this wasn't left out intentionally so I added it while I was looking.

nengo_spa/algebras/vtb_algebra.py
```diff
@@ -155,6 +218,81 @@ def invert(self, v, sidedness=ElementSidedness.TWO_SIDED):
...
+        except ImportError:
+            if int(exponent) == exponent:
+                exponent = int(exponent)
+                # Provide fallback for integer-only powers
+
+                def fractional_matrix_power(m, exp):
+                    power = np.eye(len(m))
+                    for _ in range(exp):
+                        power = np.dot(power, m)
+                    return power
```
arvoelke, line 270: I think there should be a way to write this so that it uses the base class implementation as a fallback? Also note these lines are uncovered (and same with the base class implementation).

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -102,6 +167,80 @@ def bind(self, a, b):
...
+        except ImportError:
+            if int(exponent) == exponent:
+                exponent = int(exponent)
+                # Provide fallback for integer-only powers
+
+                def fractional_matrix_power(m, exp):
+                    power = np.eye(len(m))
+                    for _ in range(exp):
+                        power = np.dot(power, m)
+                    return power
```
arvoelke, line 221: Same comment as https://github.com/nengo/nengo-spa/pull/271/files#r591639122.

nengo_spa/algebras/vtb_algebra.py
arvoelke: Pushed a commit to turn codecov on for the `no-scipy` job. Nengo appears to do the same thing. Currently we're seeing codecov patch report `84.32% of diff hit (target 90.00%)` and codecov project report `95.49% (target 90.00%)` so I'll see what it says afterward. It should go up due to `test_algebra.py::test_integer_binding_power` running without scipy, triggering the `ImportError` and then running these blocks.

arvoelke: codecov/patch: `90.29% of diff hit (target 90.00%)`
codecov/project `96.15% (target 90.00%)`

And these lines are now covered, so looks like that worked.

Remaining question from this thread:
> I think there should be a way to write this so that it uses the base class implementation instead of this fallback?

nengo_spa/algebras/base.py
```diff
@@ -220,6 +301,66 @@ def implement_binding(self, n_neurons_per_d, d, unbind_left, unbind_right):
...
+        ----------
+        v : (d,) ndarray
+            Vector to obtain the absolute vector of.
+
+        Returns
+        -------
+        (d,) ndarray
+            The absolute vector relating to the input vector.
+        """
+        return self.bind(self.sign(v).to_vector(len(v)), v)
```
arvoelke, line 362: Without thinking about it very deeply, it seems more natural to me for the default implementation for `abs` to be something like:
```
if self.sign(v).is_negative():
    v = self.bind(self.negative_identity_element(len(v)), v)
```
i.e., "'multiply' it by -1 if it is negative".

Is that consistent with this definition?

arvoelke: Similarly, would there be any use in having a `negate` method that flips the sign of `v`?

nengo_spa/algebras/tests/test_algebras.py
```diff
@@ -147,6 +195,33 @@ def test_identity_element(algebra, sidedness, rng):
...
+@pytest.mark.parametrize("sidedness", ElementSidedness)
+def test_negative_identity_element(algebra, sidedness, rng):
+    try:
+        x = next(UnitLengthVectors(16, rng=rng))
+        a = algebra.abs(x)
+        with warnings.catch_warnings(record=True) as caught_warnings:
+            warnings.simplefilter("always", DeprecationWarning)
+            p = algebra.negative_identity_element(16)
+    except NotImplementedError:
+        pass
```
arvoelke, line 207: Not sure I'm the biggest fan of these `except NotImplementedError -> pass` clauses. I think `pytest.skip()` would be a bit better. The rationale being if someone is running individual tests to debug things, or to develop their own custom algebra, then there's no way to distinguish between a test passing because it was implemented correctly versus wasn't implemented at all.

nengo_spa/algebras/base.py
```diff
@@ -288,3 +452,82 @@ def zero_element(self, d, sidedness=ElementSidedness.TWO_SIDED):
...
+    def is_zero(self):
+        """Return whether the sign neither positive nor negative (i.e. zero),
+        but definite."""
+        return not (self.is_positive() or self.is_negative() or self.is_indefinite())
+
+    def is_indefinite(self):
+        """Return whether the sign is neither positive nor negative nor zero."""
+        raise NotImplementedError()
+
+    def vector(self, d):
```
arvoelke, line 477: Should be `to_vector`?

nengo_spa/algebras/base.py
```diff
@@ -288,3 +452,82 @@ def zero_element(self, d, sidedness=ElementSidedness.TWO_SIDED):
             Zero element.
         """
         raise NotImplementedError()
+
+
+class AbstractSign(ABC):
+    """Abstract base class for implementing signs for an algebra."""
```
arvoelke, line 458: The need for this abstraction is a bit unclear to me. It seems like the sign could more simply be represented by a value that can either be -1, 0, or 1 (as in the output of `np.sign`). Then `to_vector` would look something like:
```
if sign == +1:
    return algebra.identity_element(d)
elif sign == -1:
    return algebra.negative_identity_element(d)
elif sign == 0:
    return algebra.zero_element(d)
else:
    raise ValueError("... indeterminant ...")
```
and that would just be a method on the algebra (rather than a separate class). I'm not sure what having this as a class buys us over this alternative? 

A more specific concern I have with this abstraction is it doesn't appear to guarantee that `is_positive` and `is_negative` and `is_zero` are mutually exclusive options.. when I think they should always be?

nengo_spa/algebras/base.py
arvoelke: Oh, no this won't work because there is only one negative identity for HRR but there are 3 cases that need to be handled to make it positive based on which of the two coefficients are -1 (or both).

nengo_spa/algebras/base.py
arvoelke: After thinking a bit more I realized this doesn't work for the HRR algebra since `to_vector` needs to handle 3 negative cases in order to get `abs` to work properly. But if we take a step back, `to_vector` is only needed by `abs` and `create_vector`, correct? So couldn't all that casework be the responsibility of those methods? And then you wouldn't really need the sign for anything except to limit the `binding_power` method?

nengo_spa/algebras/hrr_algebra.py
```diff
@@ -62,6 +67,45 @@ def is_valid_dimensionality(self, d):
...
+
+        if rng is None:
+            rng = np.random.RandomState()
+
+        v = rng.randn(d)
+        v /= np.linalg.norm(v)
+
+        if HrrProperties.POSITIVE in properties:
+            properties.remove(HrrProperties.POSITIVE)
+            v = self.bind(self.sign(v).to_vector(d), v)
```
arvoelke, line 98: I think this can call out to `abs` since the default implementation for `abs` is exactly this?

nengo_spa/algebras/hrr_algebra.py
```diff
@@ -82,6 +126,48 @@ def bind(self, a, b):
...
+
+        * an exponent of -1 will return the inverse,
+        * an exponent of 0 will return the identity vector,
+        * and an *exponent* of w1cne will return *v* itself.
+
+        The following relations hold for integer exponents, and for unitary
+        vectors:
+
+        * :math:`\mathcal{B}(v^a, v^b) = v^{a+b}`,
+        * :math:`(v^a)^b = v^{ab}`.
```
arvoelke, line 145: Could mention the first one holds for real exponents too. :)

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -204,3 +375,26 @@ def zero_element(self, d, sidedness=ElementSidedness.TWO_SIDED):
...
+
+
+class TvtbSign(GenericSign):
+    def to_vector(self, d):
+        if self.sign is None:
+            raise NotImplementedError(
+                "There is no vector corresponding to an indefinite sign."
+            )
+        elif abs(self.sign) > 0:
+            return self.sign * TvtbAlgebra().identity_element(d)
```
arvoelke, line 387: Since the two options are `+1` and `-1` it would be clearer IMO to call `identity_element` if it is `+1` and `negative_identity_element` if it is `-1`. This would be a bit more code, but it would deduplicate `negative_identity_element` in a sense.

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -81,6 +88,64 @@ def _get_sub_d(self, d):
             raise ValueError("Vector dimensionality must be a square number.")
         return sub_d
 
+    def create_vector(self, d, properties, *, rng=None):
```
arvoelke, line 91: Looks like this method fully duplicates VTB except for the changes in string names and the sidedness of the `identity_element`? Is there a rationale for not reusing code here (and similar elsewhere)?

nengo_spa/algebras/vtb_algebra.py
```diff
@@ -155,6 +218,81 @@ def invert(self, v, sidedness=ElementSidedness.TWO_SIDED):
...
+        of *v* to itself.
+
+        Fractional binding powers are supported for "positive" vectors if SciPy
+        is available.
+
+        Note the following special exponents:
+
+        * an exponent of -1 will return the inverse,
+        * an exponent of 0 will return the identity vector,
+        * and an *exponent* of w1cne will return *v* itself.
```
arvoelke, line 234: There is a typo of 'wicne' here versus `1` in the base docstring.

nengo_spa/algebras/vtb_algebra.py
```diff
@@ -155,6 +218,81 @@ def invert(self, v, sidedness=ElementSidedness.TWO_SIDED):
...
+            raise ValueError(
+                "Fractional binding powers are only supported for 'positive' vectors."
+            )
+
+        if exponent < 0:
+            v = self.invert(v, sidedness=ElementSidedness.RIGHT)
+
+        sub_d = self._get_sub_d(len(v))
+        power = fractional_matrix_power(
+            v.reshape((sub_d, sub_d)) * np.sqrt(sub_d), abs(exponent) - 1
```
arvoelke, line 290: Would there be any possibility for problem here if `0 < abs(exponent) < 1`?

arvoelke (changes requested):
Finished a primary round of reviewing and left a few comments. Apologies that some of them involve threads of me figuring things out incrementally as I went. Overall I think it's great and would be ready to merge in after addressing any necessary questions and comments.

I think any TODOs that are still remaining after this could also be done in future PRs.

One of my threads is wondering about how necessary the `AbstractSign` class is and if it could be removed and basically integrated into the internal logic of `abs` and `binding_power`. On the one hand I do like how this naturally explains to the user how these objects behave mathematically, but on the other hand it might be an unnecessary abstraction that makes it confusing to know where certain responsibilities lie. As a case for this point the logic that makes an HRR vector positive is kind of shared between `abs` and `HrrSign.to_vector`. Similarly `VtbSign.to_vector` kind of duplicates the logic for `negative_identity_element`. As a user the only thing I think I would care about is making a vector positive and checking if it's positive. These are really minor, all things considered, but are flags to me that the extra abstraction is being made before it's really needed.

This is quite a bit outside the scope of this PR, but while I'm thinking about it I have been wondering how this interface could work efficiently with batches of vectors and multiple consecutive bindings (being done offline without neural networks). I have been doing this quite a bit lately to prepare datasets comprised of [S]SPs and found that it's much faster (e.g., 3-5x faster) to apply just a single `rfft` to an entire batch of vectors, do all of the bindings that are needed while staying in the Fourier domain, and then applying a single `irfft` to the batch of vectors only as a final step. I don't really see a way to reuse this code in this context, but it would be nice to have that sort of capability centralized here without replicating parts of the HRR algebra code, so it's maybe something to think about.

nengo_spa/algebras/base.py
```diff
@@ -220,6 +301,66 @@ def implement_binding(self, n_neurons_per_d, d, unbind_left, unbind_right):
...
+            The sign of the input vector.
+
+        See Also
+        --------
+        AbstractAlgebra.abs
+        """
+        raise NotImplementedError()
+
+    def abs(self, v):
+        """Returns the absolute vector of *v* defined by the algebra.
```
arvoelke, line 339: One more thought: since the property `UNITARY` has the method `make_unitary`, maybe the property `POSITIVE` could have the method `make_positive` (instead of `abs`)?

nengo_spa/algebras/tvtb_algebra.py
```diff
@@ -124,7 +124,12 @@ def create_vector(self, d, properties, *, rng=None):
             properties.remove(TvtbProperties.UNITARY)
             v = self.make_unitary(rng.randn(d))
         elif TvtbProperties.POSITIVE in properties:
-            from scipy.linalg import sqrtm
+            try:
+                from scipy.linalg import sqrtm
+            except ImportError:
+                raise ValueError(
```
jgosmann, line 130: Not sure that `ValueError` is the right exception type to use here ...
(But definitely nothing wrong with providing the more precise error message.)

arvoelke: Right I was doing this to mirror https://github.com/nengo/nengo-spa/pull/271/files#diff-db444a08d43b9a69e4a19456a46ae35748e19b5e058b0e12a2955abda81eb3d3R225 which is also triggered by not having SciPy installed. I can see the reason for them being different, since the other situation is like a value error inside of an import error. I could also see them both being `ImportError` if you think of the import happening because it's fractional (even though the code has it the other way around).

arvoelke: The analogy would be the value is the "positive" property in this case versus the fractional exponent value in the other case. Both exceptions require a particular value combined with SciPy being missing.

nengo_spa/algebras/vtb_algebra.py
jgosmann: I can see the desire to reduce the code duplication, but a few points kept me from trying to hard in this instance:

* It is not entirely clear to me that we have actual semantic code duplication and not just syntactic duplication. In a sense VTB and TVTB should be handled as independent algebras. (Even though they are based on the same idea, which that might suggest that there might be a reasonable common basis to extract.)
* Extracting duplicated code adds some indirection/more complex code which might make it harder to understand each of these algebras.
* [Rule of three](https://en.wikipedia.org/wiki/Rule_of_three_(computer_programming))

nengo_spa/algebras/tests/test_algebras.py
jgosmann: That's a good point and I'll consider it. But I think skipped tests are listed in the summary statistics of pytest and it might be unclear whether the tests were skipped because some flag/dependency/etc. was missing or because they are always skipped because an algebra does not implement in the feature. Maybe `xfail` would be more appropriate (because an algebra not implementing the feature is expected to fail the test), but not entirely sure. I feel it is still a slightly different usage than what `xfail` is usually used for.

nengo_spa/algebras/tests/test_algebras.py
```diff
@@ -83,6 +84,23 @@ def test_integer_binding_power(algebra, d, rng):
     assert np.allclose(algebra.binding_power(v, 3), algebra.bind(algebra.bind(v, v), v))
 
 
+@pytest.mark.parametrize("d", [25, 36])
+def test_integer_binding_base_implementation(algebra, d, rng):
```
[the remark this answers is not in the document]
arvoelke: Heh yeah I was going to leave a comment saying that this is only testing that they match, and then the previous unit test is checking that it is correct -- but then I noticed that none of the unit tests had comments so I felt awkward breaking that pattern.

nengo_spa/algebras/base.py
jgosmann: * [ ] Yes, I think `HrrAlgebra.create_vector` could just call `abs`. (For the other algebras, this is not sufficient because the sign is usually indefinite.)

* `is_positive`, `is_negative` etc. should indeed be mutually exclusive. Currently, it is up to the implementer to ensure this. I agree, that it might be preferable to ensure this on the API level, but currently I am not sure how to do that while still preserving a nice to use API. I guess, in some way the sign needs to be mapped to an enum. But taking a `sign` and then calling a function to get the "actual" sign value seems not optimal to me. I will think more about this.
* As for why I implemented this abstraction:
** I think it clearly defines one relevant aspect of the vectors. That helps (at least me) to reason about them and to implement other methods such as `abs` and `create_vector` on a higher level of abstraction.
** I also consider this part of the public API, i.e. allowing a user to check the sign of a (maybe randomly generated) vector.
** I also don't really want to include `to_vector` within `create_vector` or `abs` because it would add quite a bit of complexity to those methods while it could actually be nicely factored out (as done right now) with well defined inputs and outputs for `to_vector` and it keeps `create_vector`/`abs` at one level of abstraction.
** The `to_vector` makes also sense on a conceptual level to me because it both the sign can either be represented as one or two numbers or equivalently by the vector. The sign of a number is also a number, analogously the sign of a SPA can be a SPA.

nengo_spa/algebras/hrr_algebra.py
```diff
@@ -82,6 +126,48 @@ def bind(self, a, b):
...
+        r"""Returns the binding power of *v* using the *exponent*.
+
+        The binding power is defined as binding (*exponent*-1) times bindings
+        of *v* to itself. Fractional binding powers are supported.
+
+        Note the following special exponents:
+
+        * an exponent of -1 will return the inverse,
+        * an exponent of 0 will return the identity vector,
+        * and an *exponent* of w1cne will return *v* itself.
```
jgosmann, line 139: * [ ] Fix typo in this line.

nengo_spa/algebras/hrr_algebra.py
jgosmann: * [ ] Not obvious to me right now, but will double check that this is true and add the mention (especially because it might be non-obvious).

nengo_spa/algebras/tvtb_algebra.py
jgosmann: https://github.com/nengo/nengo-spa/pull/271#discussion_r593914636

nengo_spa/algebras/vtb_algebra.py
jgosmann: * [ ] If my math is correct that should not be a problem, but I will double check.

nengo_spa/algebras/base.py
jgosmann: Essentially for the reasons described in the PR description under "Vector creation API". It does not really scale well to add a new `make_something` method for any potential property (and potentially the cartesian product of these properties). Also, `abs` is more analogous to other APIs like `np.abs`.

nengo_spa/algebras/tvtb_algebra.py
jgosmann: I see, the reference is definitely relevant and needs to be considered together with this.

nengo_spa/algebras/base.py
arvoelke: Yeah I read that part and this isn't really about scaling with the cartesian product. It's more that I'm not sure of that analogy. For instance `np.abs(-3 - 4j) = 5` but the HRR abs as it is defined here only changes the sign of 1 or 2 coefficients while leaving the rest unchanged. So I'm not sure if this analogy helps avoid confusion whereas `POSITIVE` is nice because it is more clearly referring to a property that is defined by the algebra.

arvoelke: I was thinking that if you could implement it like:
```
def abs(self, v):
    if self.sign(v).is_negative():
        v = self.negate(v)
    return v
```
where
```
def negate(self, v):
    return self.bind(self.negative_identity_element(len(v)), v)
```
then I would see this as indeed being very much analogous to `abs`. I tried this out and two of the HRR unit tests failed. Not sure why or if that's fixable without any negative consequences? I don't think so because the vector that needs to be binded with `v` is context-sensitive, in that there are 3 different choices corresponding to the ones in `HRRSign.to_vector`.

arvoelke: Actually scratch that, I don't think my last post is a good argument at all because you could fix `negate` by defining the LHS term in a way that is context sensitive in `v`, and there's no good reason to require a priori that `abs` should be defined in this way. The example I gave of `np.abs(-3 - 4j) == 5` is an example where `abs` cannot be defined in such a way.

Perhaps the part that is making me uncomfortable is there's generally only one way to flip the sign of something, and I believe that is usually consistent with multiplying by the negative identity. But here there are two ways to make something negative. And to make something positive, the thing that you have to multiply it with varies depending on which way it is negative. So I guess I'm having a hard time making these analogies fully consistent with one another in my head right now.

nengo_spa/algebras/base.py
```diff
@@ -220,6 +301,66 @@ def implement_binding(self, n_neurons_per_d, d, unbind_left, unbind_right):
...
+        """Returns the sign of *v* defined by the algebra.
+
+        The exact definition of the sign depends on the concrete algebra, but
+        should be analogous to the sign of a number in so far that
+
+        * binding of two "positive" vectors produces another "positive" vector,
+        * binding of a "negative" vector with itself produces a "positive"
+          vector,
+        * binding of a "negative" and "positive" vectors produces a "negative"
+          vector.
```
arvoelke, line 314: After writing https://github.com/nengo/nengo-spa/pull/271/files#r595699304 I realized this might mislead people as one might expect that binding one "negative" vector with another "negatve" vector would be guaranteed to produce a "positive" vector. However, in the HRR algebra, with even dimensionality, there are three ways that a vector can be "negative" based on the combination of the two coefficients: `A = (-1, 1)`, `B = (1, -1)`, and `C = (-1, -1)`. Then we have:

 `*` | A | B | C |
 --- | --- | --- | ---
 A | `+` | `-` | `-`
 B | `-` | `+` | `-` 
 C | `-` | `-` | `+` 

so the analogy breaks if you are not in the same sub-group for the "negative" times "negative" case. The docstring says "with itself" so that is okay but it might still defy people's expectations.

Complex numbers solve this by not reducing the sign to "positive" or "negative" but keeping it as `z / abs(z)` (keeping phase and throwing away amplitude). Just as we are keeping it as this two-tuple and throwing away all the other coefficients. Then `sign(X*Y) = sign(X)*sign(Y)` is the sort of thing that I think a user can expect to hold in general, as it does here.

So I guess my suggestion then is to say here that it is analogous because of `sign(X*Y) = sign(X)*sign(Y)` and then for the HRR sign, you can define what it means to mulyiply `sign(X)` and `sign(Y)` by multiplying those tuples element-wise with one another.

nengo_spa/algebras/base.py
arvoelke: Okay I think I've figured out what was making me uncomfortable and how to fix it. I see the analogy to `abs` now as being derived from this definition of sign:
```
sign(z) = z / abs(z)
```
which is exact same definition as the one for complex numbers, and could also be the definition here. Indeed, if you multiply both sides by `abs(z)` then you have:
```
abs(z) = (1 / sign(z)) * z
```
where `1 / sign(z)` is consistent with `HRRSign.to_vector` because `1 / -1 == -1 ` and `1 / 1 == 1`. And `*` is binding. This is currently the base implementation for `abs`! Hurray.

Somewhat different from how I had been thinking about this until now, but I think that actually makes the analogy more complete and consistent in combination with [the comment about multiplying signs](https://github.com/nengo/nengo-spa/pull/271/files#r595706491)...?

nengo_spa/algebras/hrr_algebra.py
jgosmann: Given non-unitary vectors, this actually only holds for **non-negative** real exponents. A negative exponent implies an approximate inverse (at least within Nengo SPA) and that invalidates the relationship.

nengo_spa/algebras/base.py
jgosmann: I wasn't even aware of the `sign` definition for complex numbers, but it seems to fit nicely.

nengo_spa/algebras/base.py
jgosmann: * [ ] This is a good point that should be added do the documentation.

I will also have a look whether some general documentation/user guide besides the bare API documentation should be added.

nengo_spa/algebras/tests/test_algebras.py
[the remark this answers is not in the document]
jgosmann: Ah ok, I didn't see the other test because it was cut off in the diff. I added a commit to slightly rename the test to express that it is only intended to test consistency between the concrete algebra and the base implementation. Together with the other test this should be sufficient to test the base implementation. Thus, I wouldn't go through the lengths to test the base implementation in isolation by either using a mock (which would likely coupled to the implementation) or dummy algebra (too much effort for a single test case).

jgosmann:
I addressed all minor comments. With regards to the major comments:

* The sign abstraction and code duplication between `VtbAlgebra` and `TvtbAlgebra`: I added comments with my reasons for the design decisions. If that convinces you, great! If not, feel free to poke me again about and I will spend some more thought on it as I think that I at least roughly understand your angle.
* [ ] The `ValueError` vs `ImportError` I will still have to take a careful look at.
* More documentation needs to be added (this is also marked as todo at top of this PR). This will involve some of the sign aspects you commented on.
* I will leave `except NotImplementedError -> pass` as is in this PR to keep it focused on the binding powers. Though, after thinking a bit about this, I think that `xfail` would be correct approach here. However, this is not completely trivial. Currently, the `NotImplementedError` is used to detect if a feature is unsupported. When using `xfail` this should be known a-priori (so that we can differentiate between "not **yet** implemented" and "intentionally not implemented/unsupported"). This might involve adding a way for algebras to declare their "supported features".

nengo_spa/algebras/tvtb_algebra.py
jgosmann: I decided to use `ImportError` (or `ModuleNotFoundError` for Python >=3.6) in both places. While one could argue that the value is invalid given an environment without SciPy, it still comes down to not being able to import SciPy and if SciPy were available, the value would be valid. So the fix is not to fix the value, but provide SciPy.

jgosmann:
Some thoughts on one of the remaining points: A Fourier coefficient might have a complex angle close to π which might be somewhat "randomly" either >π or <π due to limitation in floating point precision (not sure what exactly the numerical stability properties of the FFT are). Whether the angle is >π or <π determines the direction of the phase shift during exponentiation. Thus the fractional binding power might be influenced by this. The question is should we concerned about this?

* It will rarely happen, but every now and then an angle will be close to π.
* It will only affect one (or a few) coefficients, thus having limited impact on the final result.
* One might no even care. Usually, the vectors are generated random anyways and the direction of phase shift does not really matter as long as its consistent.

For these reasons, I would definitely not raise a hard error by default. Instead I would make the error treatment configurable similar to handling of floating point errors in NumPy. One would essentially get three modes: `ignore`, `warn`, and `error` (with presumably `warn` being the default). However, this adds some complexity, and I'm not sure whether that is actually required. (Besides what would be the right threshold to trigger the error?).

jgosmann:
@arvoelke I think this PR is ready for another review round. From my perspective the only one open point left, see my previous comment. Some remaining work, to make binding powers ready for release (`SemanticPointer` API, vector generators, user guide documentation), I will do in follow-up PRs (e.g. #282).

The Travis-CI failure should be fixed by #283.

jgosmann:
I noticed that something with the HRR sign does not seem to add up: The vector `v = np.array([0., 1., 0., 0.])` should be positive because it only shifts another vector under binding, however:

```python
spa.algebras.HrrAlgebra().sign(v)
# HrrSign(dc_sign=1, nyquist_sign=-1)
```

This seems to contradict @bjkomer thesis:

> Only one of these disjoint spaces is useful for SSPs, the one in which the constant and Nyquist coefficients are positive. If one of those components is negative, each application of self convolution will flip the sign, causing the result to jump back and forth between two regions. This becomes a problem when fractional binding is used, as partial convolutions will leave the result between the two spaces (the resulting vector will have both real and imaginary parts, and when the imaginary part is discarded, it will no longer be unitary).

🤔

arvoelke:
I'll try to get back to this this week. Just looking at your most recent comment, I'm not sure if I see the contradiction? That is I'm not sure why `v = np.array([0., 1., 0., 0.])` should necessarily be positive. It does shift another vector under _integer_ binding, but why does that mean it must be positive?

To me the relevant question is if `v ** exp`, where `exp` is non-integer, is guaranteed to be real. Here's a quick snippet that shows that it is _not_ guaranteed to be real:

```python
import numpy as np

import nengo_spa as spa

d = 2

algebra = spa.algebras.HrrAlgebra()
v = np.roll(algebra.identity_element(d), 1)

print(np.fft.ifft(np.fft.fft(v) ** 0.5))
```
=> `[0.5+0.5j 0.5-0.5j]`

That makes sense since the Nyquist coefficient undergoes `\sqrt{-1} = \pm i` and then the inverse Fourier transform of either choice results in imaginary coefficients. This is summarized by the fact that a signal is real if and only if its DC and Nyquist term (if applicable) are real. I'm not of a citation for this fact off-hand, but I think it has something to do with each needing to be its own complex conjugate (that is, `x == np.conj(x)` if and only if `x` is real).

If you remove the validation in `binding_power`, you'll see that `np.fft.irfft` is assuming the output must be real and throws away the imaginary component.
```python
print(algebra.binding_power(v, 0.5)) 
```
=> `[0.5 0.5]`

Therefore it seems consistent to me that `v` is not positive and is not allowed to be used as the base vector for a fractional binding power. It would be okay if the result were allowed to be imaginary, which would require using `np.fft.[i]rfft` instead of `np.fft.[i]fft`.

arvoelke:
> Some thoughts on one of the remaining points: A Fourier coefficient might have a complex angle close to π which might be somewhat "randomly" either >π or <π due to limitation in floating point precision (not sure what exactly the numerical stability properties of the FFT are). Whether the angle is >π or <π determines the direction of the phase shift during exponentiation. Thus the fractional binding power might be influenced by this. The question is should we concerned about this?
> 
> * It will rarely happen, but every now and then an angle will be close to π.
> * It will only affect one (or a few) coefficients, thus having limited impact on the final result.
> * One might no even care. Usually, the vectors are generated random anyways and the direction of phase shift does not really matter as long as its consistent.

I don't think I ever mentioned this but this was actually the very first implementation that I was using for at least one of our papers on this subject. I added a small epsilon to stay away from `+/- pi` in either direction. That implementation happens to be preserved here: https://github.com/bjkomer/spatial-semantic-pointers/blob/b6e7453dfe4a3c63aed66a9fe7ccfc2d5f013147/spatial_semantic_pointers/utils.py#L34-L55

It turns out that this was unnecessary and that I was being overly cautious. Recall that this only comes up when generating base vectors and there the numerical precision of the FFT methods seem to be just fine. At some point in time we all switched to not include this safety margin and it's been totally okay. An analogy might be if you were uniformly generating numbers in the range `[-1, 1]` and testing if that number is negative or positive. You statistically don't need to worry about the `0` case as that [almost surely](https://en.wikipedia.org/wiki/Almost_surely) never happens (technically this assumes a real domain, but on a digital computer the probability would be on the order of default floating precision). 

If you were implementing the `algebra.binding_power` method in neurons with a fixed `exponent` and variable `v` then you might want to stay away from `+/- pi` as that corresponds to a discontinuity in the function space with respect to changes in `v`, similar to `arctan`. Depending on the implementation those might be infeasible for the tuning curves to fit (e.g., see [Nengo arctan discussion](https://forum.nengo.ai/t/neural-implementation-of-arctangent2/869/4?u=arvoelke)). But I don't recall if we've needed to do this yet. Usually we do the opposite of implementing `algebra.binding_power` with a fixed `v` and a variable `exponent`, which doesn't have this same discontinuity with respect to changes in the `exponent`.

nengo_spa/algebras/hrr_algebra.py
```diff
@@ -210,3 +342,132 @@ def zero_element(self, d, sidedness=ElementSidedness.TWO_SIDED):
...
+
+        ============  =================  ====================================
+        DC component  Nyquist component  Vector
+        ============  =================  ====================================
+         1            >=0                [ 1,  0, 0, ...] (identity)
+         1             <0                [ 0,  1, 0, 0, ...]
+        -1            >=0                [ 0, -1, 0, ...]
+        -1             <0                [-1,  0, 0, 0, ...] (negative identity)
+         0              0                [ 0,  0, 0, ...] (zero)
+        ============  =================  ====================================
```
arvoelke, line 420: Last question is about this table here and about in the class docstring for `HrrSign`. Why is the DC component assumed to be in `{-1, 0, 1}` while the Nyquist component is floating? I'd think they should both be floating? If it's unitary then they would both be in `{-1, 0, 1}` but that does not need to be the case.

arvoelke (approved):
I rebased onto master to get the link update to fix Travis CI, but didn't collapse anything down yet. I added two minor fixup commits for some doc typos. I think it looks great (I also checked out the documentation and gave it a quick look and didn't notice anything else)! Just left one question about the tables in the HRR documentation. Apart from that question I think it's ready to go.

nengo_spa/algebras/hrr_algebra.py
jgosmann: Within the sign representation both are in `{-1, 0, 1}` ([see lines 387–395](https://github.com/nengo/nengo-spa/pull/271/files/ee2f7d2a776cd32dffe73af153c7dbcd917303c2#diff-daa201651cf1b7c7487c0daf2ce5b8dc327330530a1cabf826953c623bac3fe1R387-R395)). Using `>=`/`<` notation requires two rows less in the table. I thought that would be easier to read.

arvoelke: Oh I see. So am I to read `<0` as `-1` and `>=0` as `{0, 1}`? Confused me a bit because the columns in the table are also "DC component" and "Nyquist component" instead of "DC sign" and "Nyquist sign" (thinking latter would correspond to the parameters with the `{-1, 0, 1}` sign representation, and then the former are the components from which they are derived).

jgosmann: I can rename the columns.

arvoelke: Sounds good to go to me then. I like all of this much better long-term than what I was doing with the "nondegenerate" stuff in #243. Awesome stuff!

PR #274, opened 2020-12-01
Drop Python 3.5 support

jgosmann:
**Motivation and context:**

Drop support for Python 3.5 because nengo-bones does not work with it anymore.

* Change string interpolations and string `format()` calls to f-strings which were introduced with Python 3.6. A few `format()` calls remain where an f-string either cannot be used or would be less legible.

Fixes build for all other PRs. Other PRs should switch over to f-strings.

**How has this been tested?**

existing tests

Changes to the .nengobones.yml might be a good place?

**Types of changes:**

- Maintenance

**Checklist:**

**Still to do:**

* [x] Double check that missing string interpolation arguments have been filled in correctly.


Seanny123 (approved):

.nengobones.yml
```diff
@@ -12,26 +12,27 @@ license_rst: {}
 
 manifest_in:
   global-exclude:
-    - "*.swp"  # exclude VIM swp files
+    - "*.swp" # exclude VIM swp files
 
 travis_yml:
-  python: 3.7
+  python: 3.9
```
hunse, line 18: We don't officially support Python 3.9 yet in Nengo, though that's not to say it might not all work fine.

hunse (approved):

.nengobones.yml
tbekolay: Unofficially, I'm currently using Python 3.9, but with 3.8 the issue was Windows specific so who knows.

jgosmann:
Force pushed a few more changes:
* Added 16fb5a66ee33104554c92c781b2744e2fbc2f38e
* Amended b427ddbdd2a2fbaa9d6d486d0b972d76953ce369 to fix docs build
* Added commit to fix deprecated intersphinx links

.nengobones.yml
jgosmann: I think it doesn't hurt to run the tests on 3.9 too as long as they pass. I also think that the "Python 3.6+" formulation in the the readme is fine (NengoSPA is more platform independent than Nengo, so it is even more unlikely that a new Python release breaks things). Might consider taking the classifier out of the setup.py ... but that I would have to re-add it later.

PR #283, opened 2021-04-18
Update links

jgosmann:
**Motivation and context:**
Some links have been redirecting and the MIT press journal link was dead.

In addtion, the MIT press journal links need to be ignored because the server returns a 403 FORBIDDEN for the automated link checking request.

Touches some documentation relevant for #271.

**How has this been tested?**
Sphinx linkcheck

**Types of changes:**

**Checklist:**

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -58,6 +58,7 @@ linkcheck_ignore = [
     r"^https?://www\.amazon\..*$",
     r"^https://uwspace\.uwaterloo\.ca/.*$",
     r"^https?://journals\.plos\.org/.*$",
+    r"^https?://direct.mit.edu/neco/article/.*$",
 ]
 
 # -- doctest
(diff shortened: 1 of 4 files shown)


Seanny123 (approved):

arvoelke:
Note https://doi.org/10.1162/neco_a_01179 might be a more future-compatible version of the article link for future reference.

jgosmann:
That URL redirects to the other one, not sure if the ignore regex would work then.

arvoelke:
> That URL redirects to the other one, not sure if the ignore regex would work then.

For me it currently redirects to https://direct.mit.edu/neco/article/31/5/849/8469/Vector-Derived-Transformation-Binding-An-Improved which is the same new destination that this PR changed the link to. Does it redirect you elsewhere?

jgosmann:
No, it redirects to the same URL, but the `linkcheck_ignore` regex seems to be only checked against the first URL and not the redirection target. Ignoring all of `https://doi.org/` does not seem like the best solution either. One could match the exact URL `https://doi.org/10.1162/neco_a_01179` ... I think either variant could be a little bit of a maintenance burden at some (either due to a long list of single papers in `linkcheck_ignore` or because the Neco URLs keep changing). I figured even if the link breaks again, the CI will alarm me/us and the reference is complete enough to google it.

PR #289, opened 2021-07-06
Fix NumPy deprecation warning

jgosmann:
**Motivation and context:**

See [1] for details in the release notes. Using `np.floating` will
capture more types than np.float, but is arguably closer to what one
would expect from an `is_float` function and consistent with the
`is_integer` function.

[1]: https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated

none

**How has this been tested?**

existing tests

**Types of changes:**

**Checklist:**

I think this change does not need a changelog entry as it shouldn't affect the public API, but if desired I can add one.


tbekolay:
I think `np.floating` is not meant to be used this way; a few things I tried in the terminal:

```
>>> isinstance(1.0, np.floating)
False
>>> isinstance(np.array(1.0), np.floating)
False
>>> isinstance(np.array([1.0]), np.floating)
False
```

Probably to make this work we'll need to check if it is indeed a NumPy array first, then check the dtype; e.g.

```
>>> np.issubdtype(np.array([1.0]).dtype, np.floating)
True
```

This might do the same thing, and seems a bit more obvious:

```
>>> np.array([1.0]).dtype.kind == "f"
True
```

jgosmann:
I updated the PR. This probably needs to be fixed in Nengo core too.

jgosmann:
Finally got a green build. :)

Seanny123 (approved):

PR #291, opened 2021-11-06
Use xfail instead of silently passing tests

### LINKED ISSUE #281: Use xfail instead of NotImplementedError to skip tests unsupported by an algebra
[See this comment.](https://github.com/nengo/nengo-spa/pull/271#issuecomment-801393260)

> I will leave `except NotImplementedError -> pass` as is in this PR to keep it focused on the binding powers. Though, after thinking a bit about this, I think that `xfail` would be correct approach here. However, this is not completely trivial. Currently, the `NotImplementedError` is used to detect if a feature is unsupported. When using `xfail` this should be known a-priori (so that we can differentiate between "not **yet** implemented" and "intentionally not implemented/unsupported"). This might involve adding a way for algebras to declare their "supported features".

jgosmann:
**Motivation and context:**
This makes it more clear when certain features are not supported by an
algebra.

To be able to use xfail, this adds a mechanism to declare what sidedness
values are supported by specfic algebra operations. This mechanism is
a decorator which adds an attribute with the allowed sidedness values to
the method object.

Closes #281.

none

**How has this been tested?**
existing tests

**Types of changes:**

**Checklist:**


jgosmann:
@bjkomer, @arvoelke Does anyone have time to review this soon? Then it could be included in 1.3 release. (No worries if not, then it will be released later. It is barely a user-facing change anyways.)

PR #296, opened 2025-05-30
V2.0.1.dev0 req upper limits

pmfurlong:
**Motivation and context:**
This PR places an upper limit on the packages required for nengo spa 2.0.1

**How has this been tested?**
Performed a clean install of the software packages and ran simple scripts to ensure we did not get errors known to arise from version mismatch.

This is a quick PR.

review the setup.py file

**Types of changes:**

**Checklist:**

- [ X ] I have read the **CONTRIBUTING.rst** document.
- [ X ] I have updated the documentation accordingly.
- [ X ] I have included a changelog entry.
- [ N/A ] I have added tests to cover my changes.
- [ X ] I have run the test suite locally and all tests passed.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -19,6 +19,13 @@ Release History
    - Removed
    - Fixed
 
+2.0.1 (2025-05-30)
+==================
+
+**Changed**
+
+Modified setup.py to specify an upper limit on dependencies' versions.
+
 
 2.0.1 (unreleased)
 ==================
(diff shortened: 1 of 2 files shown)


jgosmann:
Note that I tried to introduce upper version constraints in #264, but was pushed back against. I still think upper bounds are a good idea.

tcstewar:
Thanks for this, and thanks for the comment @jgosmann !

I'm not entirely opposed to upper bounds, but this seems a bit too aggressive to me.  For example, this limits the Python version to 3.8 or 3.9.  We've finally got Nengo working across 3.8-3.13, so I'd prefer to also have this one work across a reasonable range, especially if the same sorts of fixes that applied to nengo could be applied here.

